PR #25, opened 2015-10-14
Extended gplay.developer functionality

yura415:
Made gplay.developer method act like gplay.list and gplay.search.

Added fullDetail and num apps to retreive options support.


facundoolano:
Awesome, thanks!

PR #31, opened 2015-10-30
Update app.js

KKitar:
Update to retrieve Apps description.


facundoolano:
Why is this needed? description was already present, and with this change it doesn't look like descriptionHTML would keep working

KKitar:
Well, I tried to use your code to retrieve description but it doesn't work so I had  to make some changes and now it works perfectly.

facundoolano:
I did a test today with the current dev code and it worked. Can you show me the code that wasn't working for you?

facundoolano:
I was able to reproduce this now, and pushed a fix. Thanks!

PR #39, opened 2015-11-09
Changes to app.js: Getting genres and screenshots

karvounis-e:
This is a change that I required for my project. There are a few games out there that have more than one genres (https://play.google.com/store/apps/details?id=com.zonbiapps.zonbisunrise or https://play.google.com/store/apps/details?id=com.popcap.pvz_row). The code below checks the length of the genreField if it is greater than 1, gets every genre, splits it with the '&' delimiter and then clear the duplicate genres from the genre array. Also, the code gets the screenshots for every game.


facundoolano:
There are several issues with this code.
- It breaks a couple of styling conventions (use single quotes, spaces after parenthesis). [This](https://github.com/facundoolano/google-play-scraper/blob/dev/.eslintrc) is what I'm using for style.
- If genre can be multiple now, it better always be an array.
- Don't go splitting the '&', if the genre is 'Action & Adventure' leave that name.
- The code needs polishing. You're looping twice, creating unnecessary variables, etc.

Also, it looks like the genre structure has changed. The 'Action & Adventure' is a subcategory of 'Family' which isn't considered in our constants. I've created #41 to track that.

facundoolano:
@karvounis do you know if multiple genres is still a thing? Your examples are single genre now, apparently you can only have a second category for FAMILY apps.

PR #48, opened 2015-11-17
Added a method to get reviews for a game

ghost:
I have been trying for the past 2 days to find a way to automate the process of fetching reviews for a specific game. I know that this code looks ugly but it is working :) . Please feel free to talk about this pull request and make it perfect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,3 +192,47 @@ Results:
   'panda run',
   'panda pop for free' ]
 ```
+### review(opts)
+Retrieves a page of reviews for a specific application. Options:
+
+* `id`: Unique application id for Google Play. (e.g. id=com.mojang.minecraftpe maps to Minecraft: Pocket Edition game).
+* `sort`: The way the reviews are going to be sorted. Accepted values are: 
+  * newest,
+  * rating,
+  * helpfulness. 
+* `reviewType` (optional, defaults to 0, min = 0, max = 1). 
+* `page` (optional, defaults to 0): Number of page that contains reviews. Every page has 40 reviews at most.
+
+Example:
+
+``` javascript
+var gplay = require('google-play-scraper');
+
+gplay.review({
+  id: 'com.mojang.minecraftpe',
+  reviewType: 0,
+  page: 0,
+  sort: 'rating'
+}).then(function(apps){
+  console.log('Retrieved ' + apps.length + ' reviews!');
+}).catch(function(e){
+  console.log('There was an error fetching the reviews!');
+});
+```
+
+Results:
+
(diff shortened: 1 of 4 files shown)


facundoolano:
Great contribution!

First thing that comes to mind: do you really need to set all those parameters or some of them are optional? For example `authuser: '0' `, `xhr: 1`, `'cache-control': 'no-cache'` are total boilerplate, it would be best to remove them if not required.

Apart from that I'd change some field names to be consistent with the style of the rest of the package:
review -> reviews
pageNum -> page
reviewSortOrder -> sort
rating -> score
hl -> lang
dateReviewPosted -> date

karvounis-e:
Cheers @facundoolano ! I have to say that this is my first try to get the reviews automatically from PlayStore. I would bet that there are things here and there in the code that can be removed as I am not sure about their impact on the POST request. I think that the parameters should stay like that as they are checked from PlayStore and it will mess its logic :) . Can it be merged into the dev branch so that others can play with it?

facundoolano:
I'd rather merge after testing and polishing. Renaming the parameters for example (not the ones internally sent to google but the options passed to the reviews method) is better done before people start using it. Also the README should have the details on how to use the different parameters (BTW the sort order should take readable values like 'newest', 'rating', etc. and internally convert to the integers google expects).

karvounis-e:
1. I think one solution would be to add the integers google expects to the constants.js file.
2. Add a validation method for the parameters that someone passes
3. Update README
4. Research what reviewType and reviewSortOrder variables take as input. If anybody can share his knowledge with that, I would be very happy.

karvounis-e:
@facundoolano Please take a look at the updated pull request and let me know.

karvounis-e:
@facundoolano Is it better now?

karvounis:
The ghost is me btw :)

facundoolano:
Looks good.

facundoolano:
@karvounis what's `reviewType`?

karvounis:
I do not have a clear idea yet. I asked in the past if anybody knew but there was no reply.

facundoolano:
Ok, it's not useful for the module user then, I hid it.

karvounis:
The problem is that reviewType is required for the POST request.

facundoolano:
I know, but there's no need to expose it as a parameter of the API if we can't explain what it does (and it doesn't look like it does anything). I just pass 0 in the POST request.

karvounis:
Ok I see what you did there. It always takes the value 0. Unless we find an explanation for that field, it is better to be hidden. Nicely done

PR #59, opened 2016-01-10
Added country parameter in app method

pingyen:
If the app is not published in US, we cannot fetch the app data by current app method, so I added parameter country in app method.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -8,11 +8,12 @@ var url = require('url');
 var h = require('./helpers');
 
 
-function app(id, lang) {
+function app(id, lang, country) {
     lang = lang || 'en';
+    country = country || 'us';
 
     return new Promise(function (resolve, reject) {
-        var reqUrl = url.parse(playStoreUrl + '?' + queryString.stringify({id: id, hl: lang}));
+        var reqUrl = url.parse(playStoreUrl + '?' + queryString.stringify({id: id, hl: lang, gl: country}));
         request(reqUrl.href)
             .then(cheerio.load, h.requestError)
             .then(parseFields)


gurdotan:
+1

facundoolano:
Can you provide a couple of URLs showing the difference of specifying the country? When I first coded this the gl parameter seemed to be ignored in the app detail.

pingyen:
Hi @facundoolano,

Sorry for late response.

You will get 404 in following URL:

https://play.google.com/store/apps/details?id=air.com.igroslot.columbusSlotsAndroid&gl=us

You can get a Russian page in following URL:

https://play.google.com/store/apps/details?id=air.com.igroslot.columbusSlotsAndroid&gl=ru

Currently, I do not have many samples, but if the app is not published in US, we need parameter country.

Thanks,
Ping-Yen

facundoolano:
I don't have a computer available to test this currently, I'll review as soon as I can

gurdotan:
@pingyen I took the liberty of testing this quickly with the above URLs but it seems that the URLs above don't work without this PR as well

pingyen:
Hi @gurdotan,

I do not get the point.

What do you mean "the URLs above don't work without this PR as well"?

If you use an Israel IP, the following URL should be OK:

https://play.google.com/store/apps/details?id=air.com.igroslot.columbusSlotsAndroid

Because app air.com.igroslot.columbusSlotsAndroid is published in Israel.

But if you use a US IP, the above URL will not work, because app air.com.igroslot.columbusSlotsAndroid is not published in US.

That is why we need this PR.

If we have this PR, we can access an Israel app by a US IP.

gurdotan:
@pingyen aah got your point, thanks.

facundoolano:
Ok @pingyen I finally got the chance to test your links above and see why this is needed. Now, before I merge this I think it makes more sense to use an opts object as in the rest of the methods of the scraper, since the parameters are now three. Also this new parameter will need documentation in the README. 

You can do that in this PR if you want otherwise I'll end up doing it myself in the next couple of days.

facundoolano:
This is solved in 9cc94f52fd891f2e7bd2760c1804a3e39f79b215

PR #60, opened 2016-01-11
Add in-app purchase and ad-supported tags

gurdotan:
See attached screenshot.  This PR collects the tags that show the app's monetization scheme (in-app purchase or ads).  Google already tags apps that support in-app purchase. It now also requires developer to tag their apps as ad-supported if advertising is being used to monetize the app: http://www.engadget.com/2015/11/19/google-play-ad-supported-app-label/

image


facundoolano:
Thanks!

PR #61, opened 2016-01-12
Add developer email and website to the app

gurdotan:
Bring in the app developer's email and website as scraped from the links at the bottom of the page. If would like, we could add an optional `options` object to the app scraping function where one could pass in which fields it ones to be scraped or not.

image


gurdotan:
By the way, generally some of the scraping code could break if a field is missing from a certain app for some reason.  That is to say, the code could be more resilient.  Have you ever confirmed that an app page on Google Play will always have all the fields we scrape?

facundoolano:
The code is supposed to work when fields are missing (unless some optional field was assumed to be mandatory). If you find a case where the code breaks please report it as an issue.

PR #62, opened 2016-01-20
Safe guard the code in case of no developer email/website info

gurdotan:
I discovered that my previous PR #61 could crash the code in case the developer email or website info was missing from the page.  This patch protects from that by checking for nulls.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -52,9 +52,9 @@ function parseFields($) {
     var maxInstalls = cleanInt(installs.split(' - ')[1]) || undefined;
 
     var developerEmail = (additionalInfo.find('.dev-link[href^="mailto:"]').attr('href') || '').match(/^mailto:(.+)$/);
-    developerEmail = developerEmail ? developerEmail[1].trim() : null;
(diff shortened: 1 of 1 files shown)


facundoolano:
I don't like using null, I think it's better to set the property as undefined instead of not setting it if missing. Also I'd like to remove those regular expressions to get the email and website, a split should do fine there.

gurdotan:
@facundoolano I changed it to contain `undefined` in case the developer website or email aren't found.  Regarding the regex, it filters out a cleaner URL than a split.  In contrast, see this split example:

```
($('.details-section-contents').find('.dev-link[href^="https://www.google"]').first().attr('href') || '').split("https://www.google.com/url?q=")

// ["", "http://supercell.helpshift.com/a/clash-of-clans/&sa=D&usg=AFQjCNFcI_matDwefmbD0Azk46RfAf9Hag"]
```

Note that if we split on the prefix of `https://www.google.com/url?q=` the resulting URL contains some more URL params that Google Play adds which aren't related to the developer's URL.  Let me know if you still prefer to split and I'll add it.

facundoolano:
Closing in favor of #65

PR #64, opened 2016-01-26
add an option to get similars apps from google play

EranGoldman:


facundoolano:
This looks pretty good, but we should aim to return a list of objects as close as possible to the app detail request (without further request obviously). Thus the results should have appId, name, etc.

facundoolano:
Added this in efcc20bcf9d4c1fd78a899dbbfa0c1faa2480980

PR #69, opened 2016-02-12
fix post /store/getreviews since it was returning 400

cscatolini:
The reviews method was returning a 400 status code with a "Not found" message. I inspected the request made by the PlayStore using the browser and found out it was missing a parameter in the POST body. I added it and the request seems to work fine again.


facundoolano:
Nice catch, thanks!

PR #70, opened 2016-02-15
Update dependencies to their latest version

ratson:


facundoolano:
For the record this broke the code because `pluck` was removed on the latest version.

I plan to add some sort of unit tests and CI to avoid this kind of errors.

PR #72, opened 2016-02-29
.throttle option to 'search' method (try to avoid 503 error)

vkfont:


lib/search.js
```diff
@@ -4,10 +4,11 @@ var cheerio = require('cheerio');
 
 var h = require('./helpers');
 
+var optsGlobal = null;
```
facundoolano: no need to use a global, just pass the throttle opt to doSearch

facundoolano: actually I think you can just put it in the query object to be consistent with the rest of the code

lib/search.js
```diff
@@ -53,8 +54,13 @@ function doSearch(query) {
         query.apps = query.apps.concat(apps);
         return query;
     }
-
-    return request(buildUrl(query))
+    var delay = h.calcDelay(optsGlobal.throttle, request._last);
+    return Promise
```
facundoolano: Isn't there a way to avoid defining a promise twice? Plus it looks like you are delaying before making the first request. Maybe the delaying should be handled in the checkFinished step, before making the next call to doSearch

README.md
```diff
@@ -130,6 +130,7 @@ Retrieves a list of apps that results of searching by the given term. Options:
 * `lang` (optional, defaults to `'en'`): the two letter language code used to retrieve the applications.
 * `country` (optional, defaults to `'us'`): the two letter country code used to retrieve the applications.
 * `fullDetail` (optional, defaults to `false`): if `true`, an extra request will be made for every resulting app to fetch its full detail.
+* `throttle` (optional, defaults to `0`): delay between API requests (to avoid 503 error).
```
facundoolano: I'd call the parameter delay and explain that is useful to avoid throttling

facundoolano:
Thanks for the PR, I left a couple of comments in the code.

I wonder if the delay option should be used in other methods too, namely list.

lib/search.js
vkfont: there are two `doSearch` calls in this module - was not able to find easy way to pass throttle opt to second one without big changes, so left global (actually its global only for this module env)

why we should pass it into query? no need to send `throttle` to google api :)

lib/search.js
vkfont: i don't see promise defined twice, can you point it please?

if we'll handle delay _after_ request - this wont be fair delay (actual throttle minus network delay)
now we are making delay and _immediately_ call google api
my opinion: delaying little more is much better for business logic than little less

vkfont:
You're welcome

Actually, i perform some tests with delays and dint find much difference: google keep banning my requests :) So it can be useful in some cases but unfortunately not in my. Perhaps, i should thing about proxy pools.

README.md
vkfont: Yep, this will be better :)

facundoolano:
So this doesn't really prevent hitting the throttling limits?

vkfont:
Not sure, but i'll leave it in production. Anyway could be useful, isn't? :)

facundoolano:
I'll make some tests myself. If it doesn't avoid the throttling problem then it would be confusing to provide that option.

vkfont:
Any success on this, mate?

facundoolano:
FYI added a throttling option to avoid these issues: https://github.com/facundoolano/google-play-scraper#throttling

PR #73, opened 2016-03-01
Added developer's reply for each review

liranbg:


lib/reviews.js
```diff
@@ -57,15 +57,23 @@ function parseFields($) {
 		var title = reviewContent.find('span[class=review-title]').text().trim();
 		var text = filterReviewText(reviewContent.text().trim(), title.length);
 
+		var developerComment = $(this).next().first();
+		if (developerComment.attr("class") == "developer-reply")
+		{
```
facundoolano: this curly brace should be on the previous line ;) (check the style files in the root dir)

lib/reviews.js
```diff
@@ -57,15 +57,23 @@ function parseFields($) {
...
+
 		var allInfo = {
 			userId: userId,
 			userName: userName,
 			date: date,
 			score: score,
 			title: title,
-			text: text
+			text: text,
+			devDate: devDate || '',
```
facundoolano: I think `replyDate` and `replyText` are better names

lib/reviews.js
```diff
@@ -57,15 +57,23 @@ function parseFields($) {
 		var title = reviewContent.find('span[class=review-title]').text().trim();
 		var text = filterReviewText(reviewContent.text().trim(), title.length);
 
+		var developerComment = $(this).next().first();
+		if (developerComment.attr("class") == "developer-reply")
+		{
+			var devDate = developerComment.find("span.review-date").text().trim();
+			var devText = developerComment.clone().children().remove().end().text().trim();
```
facundoolano: why not just `developerComment.children().end().text().trim()` ?

lib/reviews.js
liranbg: Your'e right. old habits:)

lib/reviews.js
liranbg: Actually it doesn't matter.

lib/reviews.js
liranbg: Changed.

lib/reviews.js
```diff
@@ -57,15 +57,23 @@ function parseFields($) {
 		var title = reviewContent.find('span[class=review-title]').text().trim();
 		var text = filterReviewText(reviewContent.text().trim(), title.length);
 
+		var developerComment = $(this).next().first();
```
facundoolano:
```
var developerComment = $(this).next('.developer-reply')
if (developerComment.length) {
```

facundoolano:
Left a couple of comments (I noticed you just committed again). Also, please update the README example to show this new field.

lib/reviews.js
liranbg: Why not .first()?
otherwise it returns with all replies and I think we wish only its relevant reply

facundoolano: You're right. My point was to filter for class in next instead of comparing it as an attribute.

facundoolano:
I've just added this, thanks for the PR!

PR #75, opened 2016-03-22
Fixing reviews title and text parsing

MatiMenich:
Google apparently inserted a new class to the "review-body" div. 

Now its 
```html
<div class="review-body with-review-wrapper">...
``` 

So the current find criteria `div[class=review-body]` wasn't able to find the div, and the review's title and text were not showing up.

I changed it to a broader criteria `.review-body` so now it's working again


facundoolano:
thanks!

PR #78, opened 2016-04-27
Add version code and apk file !

sinarahimpoor:


facundoolano:
Doesn't seem to do what the title says. If you want to add proxy you'll have add it to every method and document it in the readme.

PR #83, opened 2016-05-11
fix search.num + append according test

vkfont:
looks like google page output is changed so we need to fix GAE regexp

but, it's possible country-internalization bug (my host is from Russia), could you please check from english countriy according implemented test? thank you

### DIFF
--- a/lib/search.js
+++ b/lib/search.js
@@ -61,7 +61,8 @@ function doSearch (query) {
 function getNextToken (html) {
   // extract the token for the next page request
   // for the record, I hate regexps
-  const s = html.match(/\\42(GAE.+?)\\42/);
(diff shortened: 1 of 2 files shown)


facundoolano:
On a quick check the search seems to be working. Could you give more details about what you are trying and what's the error you are seeing?

vkfont:
There was a bug: option .num in .search method is ignored - always returns only first page
After quick research i've found what your code can't fetch token for the next page (looks like google play market html was updated and regexp became incorrect)

I've fixed regexp and append corresponding test (to avoid this bug in the future).

You're welcome :)

facundoolano:
Ok, I get it now. I used your test against the dev branch and it fails. Its passing in travis so it looks like you've fixed it, mergin now. Thanks!

PR #98, opened 2016-07-14
Fix issue with French number of installs

tgalopin:
I found a problem on minInstalls and maxInstalls number in French: the separator for the two values is not a hyphen but a word ("et" for between ... "and" ...). See https://play.google.com/store/apps/details?id=com.airbnb.android&hl=de&gl=fr.

This fix the issue for French, but I think it's worth trying to find other languages where the problem could exist.


facundoolano:
Great, thanks.

PR #99, opened 2016-07-15
Fix reviews' sort

lapastillaroja:
Current Google Play reviews' constant HELPFULNESS is `2` and not `4` so I fixed.
Also `convertSort` in `reviews.js` was redundant and buggy since it was forcing the sort option to be always `NEWEST`

### DIFF
--- a/lib/reviews.js
+++ b/lib/reviews.js
@@ -11,15 +11,13 @@ function reviews (opts) {
   return new Promise(function (resolve, reject) {
     validate(opts);
 
-    const sort = convertSort(opts.sort);
-
     const options = {
       method: 'POST',
       uri: 'https://play.google.com/store/getreviews',
       form: {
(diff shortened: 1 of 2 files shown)


lib/reviews.js
```diff
@@ -11,15 +11,13 @@ function reviews (opts) {
...
-    const sort = convertSort(opts.sort);
-
     const options = {
       method: 'POST',
       uri: 'https://play.google.com/store/getreviews',
       form: {
         pageNum: opts.page || 0,
         id: opts.appId || opts.id,
-        reviewSortOrder: sort,
+        reviewSortOrder: opts.sort,
```
facundoolano: shouldn't there be a default here?

lapastillaroja: yeah, you are right
sorry, I notice this now and seems it's already done in `dev` branch
thanks :)

PR #100, opened 2016-07-15
Fix common languages parsing and add tests

tgalopin:
This PR adds tests for a lot of common languages and countries in order to ensure parsing works properly. Some issues appeared and I fixed them here too.


lib/app.js
```diff
@@ -70,11 +70,22 @@ function parseFields ($) {
   const downloads = additionalInfo.find('div.content[itemprop="numDownloads"]').text().trim();
   let installs = downloads.split(' - ');
```
facundoolano: can you move all these split / length checks to a helper function?

facundoolano:
Looks good, I left a comment about factoring the installs split logic into a function. Fix that and I'll merge.

Thanks!

tgalopin:
Updated

facundoolano:
thanks@

PR #101, opened 2016-07-15
Normalize Android required version (only version or VARY)

tgalopin:
This PR normalizes the returned Android version to avoid having to deal with localized strings.

I checked manually and the format of versions is the same for every language (at least every language in the list I put in the tests): `<version> <localized string>` (here is the list for Android 2.1 as a gist: https://gist.github.com/tgalopin/21d6ba9fc8802b3286eb4716b03dad30).

I added tests for What's App in order to tests this behavior (Facebook is for VARY).

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -63,7 +63,8 @@ function parseFields ($) {
   const description = additionalInfo.find('div[itemprop=description] div');
   const version = additionalInfo.find('div.content[itemprop="softwareVersion"]').text().trim();
   const updated = additionalInfo.find('div.content[itemprop="datePublished"]').text().trim();
-  const requiredAndroidVersion = additionalInfo.find('div.content[itemprop="operatingSystems"]').text().trim();
+  const androidVersionText = additionalInfo.find('div.content[itemprop="operatingSystems"]').text().trim();
+  const androidVersion = normalizeAndroidVersion(androidVersionText);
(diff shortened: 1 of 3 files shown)


facundoolano:
I'm not a fan of having all those localised strings hardcoded in the app file. Since you've confirmed that specific versions always start with a number, how about doing this:

- Replace the current `requiredAndroidVersion` field with a `androidVersionText` in case someone wants to still use the localised text for the version. (that way you don't need to check if the value is `VARY` to display it).
- Have a separated `androidVersion` field where you put the normalised version. But instead of matching all possible values, just check if it's a string that starts with a number and split on the first space.

Do you think that would work?

tgalopin:
That would work, I'll do that :) .

tgalopin:
Updated.

facundoolano:
Great, thanks. I'm not sure about making the tests dependant on FB and whatsapp. I usually use my app because I know I don't have plans to change it (and I can go update the tests right away if I ever do). But FB and whatsapp could change the required version any time and break the tests.

I guess I'll leave it for now and change to a more generic test if they start giving problems.

tgalopin:
My issue with using your app is that you probably didn't put as much translations as they did. That's the reason I used famous applications: to have as much differences in languages as possible.

PR #102, opened 2016-07-15
Add SDK API level

tgalopin:
Add a field for the SDK API level.

See https://source.android.com/source/build-numbers.html


facundoolano:
I'm not that familiar with android dev, what's the use of the SDK API level?

tgalopin:
The api level is used for compatibility : as an android developer,  you use it to declare what version you support and therefore which sdk features you are using.

facundoolano:
This feels somewhat unrelated to the Google Play scraper. The change adds a bit of clutter to the code and implies maintaining that mapping of version to api level. It sounds more sensible to keep that mapping in the clients of the library. 

I'm interested in knowing if anyone else would benefit from this, though, so I'm going to leave this PR open for a while.

tgalopin:
It seems fair :) . I'll put it in my code then.

facundoolano:
Nobody chimed in asking for this, so I'm closing it for now.

PR #104, opened 2016-07-26
Detect reviews's score in Japanese

lapastillaroja:
When fetching reviews for the Japanese language, score text has this format `5つ星のうち3つ星で評価しました`, in contrast to the default english format `Rated 3 stars out of five stars`.

Uber in English: https://play.google.com/store/apps/details?id=com.ubercab&hl=en
Uber in Japanese: https://play.google.com/store/apps/details?id=com.ubercab&hl=ja

Maybe a cleaner option would be add a `country` option as in `app`. However I didn't check if that's possible for `reviews`' api. Also I tried to make the less changes possible

### DIFF
--- a/lib/reviews.js
+++ b/lib/reviews.js
@@ -109,7 +109,7 @@ function filterScore (score, lang) {
   let regex;
   if (lang === 'ja') {
     // japanese reviews: '5つ星のうち3つ星で評価しました'
-    regex = /5つ星のうち([0-5]{1})つ星で評価しました/;
+    regex = /[0-5].*?([0-5]{1})/;
   } else {
     // default: 'Rated 3 stars out of five stars'
     regex = /([0-5]{1})/;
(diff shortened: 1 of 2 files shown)


facundoolano:
There's already a `lang` option for reviews. Can you check that option instead of matching the Japanese text? (Please let me know if that option doesn't work)

lapastillaroja:
I'll take a look

lapastillaroja:
Done. Check it please

facundoolano:
This still depends on the hardcoded japanese text, which is kind of what I wanted to avoid. But nvm, I think is better to just look for any two integers in the rating text and do a `Math.min` between the two to get the rating number, which should work for any language. I'll make that fix in the next couple of days.

facundoolano:
@lapastillaroja thanks for the commit, I ended up implementing the simpler approach I described in the previous comment.

PR #119, opened 2016-10-18
Scrape the application title correctly regardless of User Agent.

jamesbrauman:
**Context**
The type of the element node holding the application title can either be a `div` or a `h1` depending on the User Agent header set when making the request. This means that the title of the application is not collected in some circumstances.

**What's this do?**
Changes the selector to not include the element type. This correctly selects the application title in all circumstances.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -39,7 +39,7 @@ function app (opts) {
 
 function parseFields ($) {
   const detailsInfo = $('.details-info');
-  const title = detailsInfo.find('div.document-title').text().trim();
+  const title = detailsInfo.find('.document-title').text().trim();
   const developer = detailsInfo.find('span[itemprop="name"]').text();
   const summary = $('meta[name="description"]').attr('content');
 


facundoolano:
I just tried a couple of apps and the title is parsed correctly. Can you show an example where it fails?

jamesbrauman:
@facundoolano 

So I did a little more digging on this. It turns out that the Play Store will return a different response based on the User Agent header that is set.

```
$ curl -s "https://play.google.com/store/apps/details?id=com.dxco.pandavszombies&hl=en&gl=us"
[snip]<div class="document-title">[snip]
```

```
$ curl -s -A "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.143 Safari/537.36" "https://play.google.com/store/apps/details?id=com.dxco.pandavszombies&hl=en&gl=us"
[snip]<h1 class="document-title">[snip]
```

Because my application uses a custom user agent header, the scraping code fails.

facundoolano:
Are you setting the agent globally or something like that? How come it affects the scraper requests?

jamesbrauman:
@facundoolano Yes, I am proxying requests through a third party proxy that makes requests from slaves all over the world, all with different User Agent headers.

facundoolano:
I honestly don't know if the module should handle that scenario, but since it's such a harmless change I'm merging it.

PR #126, opened 2016-11-16
Updated response parsing

stanislasdrg:
Updated response parsing of permissions, to reflect Google Play's latest response.


facundoolano:
thanks!

PR #129, opened 2016-12-01
Add proxy support

chriskinsman:
Added support for specifying a proxy and passing it onto request.

### DIFF
--- a/lib/list.js
+++ b/lib/list.js
@@ -11,9 +11,16 @@ function list (opts) {
   return new Promise(function (resolve, reject) {
     opts = R.clone(opts || {});
     validate(opts);
-    const url = buildUrl(opts);
 
-    request(url, opts.throttle)
+    let options = {
+        url: buildUrl(opts)
+    };
+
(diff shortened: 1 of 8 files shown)


lib/list.js
```diff
@@ -11,9 +11,16 @@ function list (opts) {
   return new Promise(function (resolve, reject) {
     opts = R.clone(opts || {});
     validate(opts);
-    const url = buildUrl(opts);
 
-    request(url, opts.throttle)
+    let options = {
```
facundoolano: make this a const, and add a `proxy: opts.proxy` instead of using a separated if. The same applies for the resto of the files.

facundoolano:
Hi @chriskinsman, thanks for the PR, I just left you a code comment. 

Also, note the build failed in Travis because of style issues, you'll have to fix that before I can merge.

chriskinsman:
All fixed up.

facundoolano:
Thanks!

PR #136, opened 2017-01-13
Add user image to review payload

seants:
Google seems to return the image URL of the user. Might as well return that if it exists. I think I check for its existence as conservatively as possible but let me know if I'm missing something.

### DIFF
--- a/lib/reviews.js
+++ b/lib/reviews.js
@@ -38,7 +38,7 @@ function reviews (opts) {
   });
 }
 
-function getUserImage($, handle) {
+function getUserImage ($, handle) {
   const spanElem = $(handle).find('span[class=responsive-img-hdpi]').find('span')[0];
   const style = spanElem && spanElem.attribs && spanElem.attribs.style;
   // Style is in the form of background-image:url(https://lh5.googleusercontent.com/.../photo.jpg)
--- a/lib/reviews.js
+++ b/lib/reviews.js
@@ -39,9 +39,9 @@ function reviews (opts) {
 }
 
 function getUserImage ($, handle) {
-  const spanElem = $(handle).find('span[class=responsive-img-hdpi]').find('span')[0];
-  const style = spanElem && spanElem.attribs && spanElem.attribs.style;
-  // Style is in the form of background-image:url(https://lh5.googleusercontent.com/.../photo.jpg)
+  const spanElem = $(handle).find('span[class=responsive-img-hdpi] > span');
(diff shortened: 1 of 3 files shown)


lib/reviews.js
```diff
@@ -38,6 +38,14 @@ function reviews (opts) {
   });
 }
 
+function getUserImage ($, handle) {
+  const spanElem = $(handle).find('span[class=responsive-img-hdpi]').find('span')[0];
+  const style = spanElem && spanElem.attribs && spanElem.attribs.style;
+  // Style is in the form of background-image:url(https://lh5.googleusercontent.com/.../photo.jpg)
+  const match = style && style.match('url\\((.*)\\)');
```
facundoolano: Try using `.css` method instead of a regex

lib/reviews.js
```diff
@@ -38,6 +38,14 @@ function reviews (opts) {
   });
 }
 
+function getUserImage ($, handle) {
+  const spanElem = $(handle).find('span[class=responsive-img-hdpi]').find('span')[0];
```
facundoolano: Can these two `find` be collapsed into one selector?

test/lib.reviews.js
```diff
@@ -8,6 +8,7 @@ function assertValid (review) {
   assert.isString(review.id);
   assert(review.id);
   assert.isString(review.userName);
+  assert('userImage' in review);
   assert(review.userName);
```
facundoolano: Use the `assertValidUrl` method for this

seants: Are you fairly comment that there will always be a URL? I wrote the getter code defensively. Didn't want to cause flakiness in the tests but I guess it would be a good signal...let me know!

facundoolano: I'm not confident, we'd have to see what gplay does when there's no image (I don't have a computer available to do it at the moment).

If there isn't always a URL we can wrap the assert on an if. Ideally the test should break whenever the selector ceases to work.

seants: Ok...maybe for now I'll assert that it's there and if you notice any failures then we can be more defensive?

For one, I did see a review where the image url appeared to link to a generic profile picture, so I believe anonymous/no-pic reviews should still have a url.

seants:
Thanks! This was my first contribution on an external github project so apologies for not having a process in place to follow up, and thanks for your help! Very excited to see this get merged :)

facundoolano:
sure, no worries, keep up the good work!

PR #149, opened 2017-04-05
Cache clp and detect it from either tag or script on page

kblcuk:
Our tests started to fail this morning; reason is that nowadays it seems that only initial page loaded has clp token (whatever that is).
So this fixes a TODO and extracts clp only after initial request.

ALSO _sometimes_ clp can't be found out from html anymore, but is present in script part (`var curl = '...blablab..?clp\x3d`); fixed that as well and added a test case for it.

### DIFF
--- a/lib/search.js
+++ b/lib/search.js
@@ -6,7 +6,13 @@ const cheerio = require('cheerio');
 const getParseList = require('./utils/parseList');
 
 function getClp (html) {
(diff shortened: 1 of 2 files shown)


kblcuk:
@facundoolano I think this is finally done now, you'd probably want to merge it soon-ish.

facundoolano:
Thanks!

PR #152, opened 2017-04-20
Solved to-do in lib/app.js file.

icaroponce:
Solved the todo for iterate over the possible characters in downloads field, instead of copying and paste for each possible case.

Also added the vim swap files ext.( .swp, .swo ) to .gitignore file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,10 @@ coverage
 # node-waf configuration
 .lock-wscript
 
+# Vim swap files
+*.swp
+*.swo
+
 # Compiled binary addons (http://nodejs.org/api/addons.html)
 build/Release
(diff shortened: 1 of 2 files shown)


facundoolano:
I'm not very fond of a regex solution either, I was thinking in something more like

```js
const characters = [' - ', ' et ', '–', '-', '～', ' a '];
const char = characters.find((char) => downloads.split(char).length === 2);
if (char) return downloads.split(char);
throw new Error(...)
```

icaroponce:
Sure! And it's really a simpler and elegant solution!
I'm gonna change it in code.

Thank you

facundoolano:
thanks!

Issue #153, opened 2017-05-16
Developer ID field?


AlexanderMatveev:
Is it possible now or any plans about getting developer id instead of developer name?

facundoolano:
That depends of what information is available in the Google Play page that's being scraped. If the method is `app` and the developer id exists in the html then it can be added to the scraped data.

AlexanderMatveev:
@facundoolano I think this information must be added to https://github.com/facundoolano/google-play-scraper#app

What field name would it be?

facundoolano:
developerId

AlexanderMatveev:
@facundoolano is this field possible only in ```app``` method? What about ```list``` method?

facundoolano:
I don't know if this field exists at all. You can check the html in the app listing and in the search results to confirm. I'm not spending time in adding features to this project right now, but feel free to send a PR if you find that field and want to add it to the results.

facundoolano:
This should be fixed by #155 (please confirm @AlexanderMatveev)

AlexanderMatveev:
@facundoolano I decided to switch to my own parser, because it looks like there is no need for client side actions and using URL query parameters is fine. Thank you for response!

Issue #113, opened 2016-09-03
Accept numerical devId


facundoolano:
For the developer method, currently '/developer?id=<name>' is used. Add support for '/dev?id=<numerical id>'

Example:

https://play.google.com/store/apps/dev?id=6695914693084197974
https://play.google.com/store/apps/developer?id=Intellijoy%20Educational%20Games%20for%20Kids

facundoolano:
don't seem worth the effort, closing.

PR #155, opened 2017-05-23
Add developerId field

icaroponce:
This pr aims to add the field 'developerId' to the apps retrieved, since it can be a util field and was requested in [#153](https://github.com/facundoolano/google-play-scraper/issues/153).

All the methods, docs and tests were update. 

It's worth to mention the change in the `developer` method api. I've decided to replace the `devId` option for `dev` (whereas it's the developer name, and not its id). It was also necessary because Google Play seems to use different urls and id patterns for developers (See below), and each one only works with one kind of id, so I decided to touch in the way that request are being made right now.

[https://play.google.com/store/apps/dev?id=5509190841173705883](https://play.google.com/store/apps/dev?id=5509190841173705883)

and

[https://play.google.com/store/apps/developer?id=DxCo+Games](https://play.google.com/store/apps/developer?id=DxCo+Games)

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,7 @@ Results:
   reviews: 2312,
   histogram: { '1': 12, '2': 7, '3': 16, '4': 40, '5': 231 },
   description: 'Everyone in town has gone zombie.',
+  developerId: "DxCo+Games",
   descriptionHTML: 'Everyone in town has gone <b>zombie</b>.',
   developer: 'DxCo Games',
   developerEmail: 'dxcogames@gmail.com',
@@ -109,6 +110,7 @@ Results:
     appId: 'com.playappking.busrush',
     summary: 'Bus Rush is an amazing running game for Android! Start running now!',
     developer: 'Play App King',
+    developerId: '6375024885749937863',
     title: 'Bus Rush',
     icon: 'https://lh3.googleusercontent.com/R6hmyJ6ls6wskk5hHFoW02yEyJpSG36il4JBkVf-Aojb1q4ZJ9nrGsx6lwsRtnTqfA=w340',
     score: 3.9,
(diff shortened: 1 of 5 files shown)


facundoolano:
I don't see the benefit of this. I think #153 was about adding the numerical id, not the text id. This new field is pretty much the same we already have in the `developer` field (it just has a different encoding but you can use the current value as the id in the developer url and it works just the same). Plus you're introducing an unnecessary breaking change in the `developer` method.

icaroponce:
Not always the developerId is the same value we already have in the `developer` field. As I've written above, Google Play seems to have 2 types of id for developers, one in text and other in a numerical form, and each apps only has one of these possible types.

For example: `5509190841173705883` its a developerId.  Once more, look at the urls above, and you'll tell the difference).

About the "unnecessary breaking change", I believe that all fields must have a clear and precision definition. what doesn't happen in the `developer` method, since the param `devId` is absolutely not the real developer Id, but his name. And, whereas the id can be different from his name (as the example I've mentioned above, which's a numerical id), it makes the method more concise and intuitive.

Please, reconsider.

facundoolano:
I saw your url, and I knew about both types of ids before. They're not one or the other, some developers have numerical (a few selected ones that have a developer home page), but all of them have text ids:

https://play.google.com/store/apps/dev?id=5509190841173705883
https://play.google.com/store/apps/developer?id=Jam+City,+Inc.

Since Google calls them and uses them as ids, developer names can still be named ids in the module. I'm all about using the proper names, but there's no clear distinction here, and even if there was, coming up with a better name is not good enough reason to introduce a breaking change in a library depended by others.

The proper way to handle this would be to accept both types of ids in the `devId` field of the developer method, then translate numerical to text. That's what #113 was about, but I closed it because it didn't seem worth the effort.

I haven't noticed before that your new field does resolve to the numerical id in the cases where there is one. That is useful, so remove the changes to the developer method and I'll merge this. Also, if you don't mind squash your commits to keep a cleaner git history.

icaroponce:
You're correct! Thank you.  

As requested, I've removed the changes in the developer method and squashed all commits into one.

facundoolano:
Cool, thanks.

Issue #154, opened 2017-05-16
[Error: Error requesting Google Play:undefined] when passing start > 199


AlexanderMatveev:
Docs for [list](https://github.com/facundoolano/google-play-scraper#list) method say:

> start (optional, defaults to 0, max is 500): the starting index of the retrieved list.

But when I'm passing a value greater than 199, getting this error:

> [Error: Error requesting Google Play:undefined]

AlexanderMatveev:
Tried to add ```reason.response.body``` before this line https://github.com/facundoolano/google-play-scraper/blob/aed23c777ed1cfcb67b2f282c8ff9ae9c999a8d3/lib/utils/request.js#L43 and got this:

> ```<html><body><h1>429 Too Many Requests</h1><p>Please reduce your request rate.</p></body></html>\n```

facundoolano:
The issue is exactly as the error says, you're requesting google too much. Try tuning the throttling to avoid hitting it too often (this has come up several times, look around the issues for more details.)

AlexanderMatveev:
@facundoolano I disagree. When I change 200 to 199, it goes normally and looks like there is no difference how often I run the scrapper.

I followed the code to find out the URL it tries to fetch, and found this URL: https://play.google.com/store/apps/category/SOCIAL/collection/topselling_free?hl=ru&gl=ru&start=200&num=10. As you can see, it always fails, no matter of request rate. What you think?

facundoolano:
I see, you are correct. Maybe the max allowed should be updated to 199, I wonder if the same happens with every collection and country combination.

facundoolano:
Fixed by #156

PR #156, opened 2017-05-27
Fix the error in list method when passing start greater than 199

icaroponce:
This fixes the bug mentioned in #154. 

I've noticed that google play is loading the content sending `POST` requests instead `GET`, also passing the start param in the req form instead of in the querystring. 

Hope you enjoy it.

### DIFF
--- a/lib/list.js
+++ b/lib/list.js
@@ -14,6 +14,8 @@ function list (opts) {
 
     const options = {
       url: buildUrl(opts),
+      method: 'POST',
+      form: opts.form,
       proxy: opts.proxy
     };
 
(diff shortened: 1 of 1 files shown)


facundoolano:
Thanks!

AlexanderMatveev:
@icaroponce @facundoolano what is the resulting URL example?

facundoolano:
@AlexanderMatveev the url is the same, but now it's called via POST instead of GET, and the start property is passed in the body

AlexanderMatveev:
@facundoolano thanks!

PR #157, opened 2017-06-09
Fix parse html when use mocks

jarcodallo:
When use mocks for caching data, is necessary parse the html before cheerio.load


facundoolano:
I'm not following what the new code is intended to do, can you elaborate on why this is necessary?

Issue #142, opened 2017-02-17
memoization overhaul


facundoolano:
memoization by default seem to have caused some problems, which in turn are harder to adress with the current all or none behavior. So:

- [ ] disable memoization by default
- [ ] allow to enable it with sane defaults with `cache: true`
- [ ] allow to configure it with `cache: opts` where opts are passed to memoizee
- [ ] make sure the cache settings affect `fullDetail` calls (i.e. if memoization is on, the inner app detail calls should be memoized too)
- [ ] set the proper promise configuration of memoizee
- [ ] do the same in app-store-scraper if it works well
- [ ] update aso module

Issue #137, opened 2017-01-16
NodeJS doesn't exit after retrieved list


vithd:
Hi, I'm having problem after getting list. My app doesn't stop after the list is shown.

```
var gplay = require('google-play-scraper');

gplay.list({
    category: gplay.category.GAME_ACTION,
    collection: gplay.collection.TOP_FREE,
    num: 2
  })
  .then(console.log, console.log);
```

facundoolano:
this has something to do with memoization (doens't happen with `cache: false`). Looking into it.

facundoolano:
I'm thinking this may be an expected or known issue when using memoization. If the script is short lived, it shouldn't do memoization. This issue should be covered by #142

andreilima:
I solved it by adding 
`.then(result => {
  console.log(result, '\n');
  process.exit(1);
}, console.log);
`

icaroponce:
The problem seems to occur always that the maxAge is set in the config object to memoizee function. The application lasts the same time to exit as it's set on this maxAge argument. if I set maxAge to 5000, the script won't exit until this time is finished ( 5s + the real time that the operation should have took ). 

Using the v3.2.4, my memoization is still be used by default, I have to set cache: false to every method i call.

Trane9991:
Example
```
var gplay = require('google-play-scraper');

gplay.app({appId: 'com.dxco.pandavszombies'})
  .then(console.log, console.log);
``` 
doesen't return too.

facundoolano:
I removed default memoization, so this should be fixed in the latest version (4.0.0)

PR #158, opened 2017-06-09
set optional memoization

icaroponce:
Since this feature was already discussed and implemented on `app-store-scraper` module, and have seen that there're still some opened issues here related to it, I decided to port it from the `app-store-scraper` module to this one, making both of them compatible in the memoization API.

It should fix the #137 and #142 issues as well.


facundoolano:
The problem with this is that in this case the utils/parseList.js file is using internally the app.js code (for the fullDetail calls), and memoizing options in this PR are not applied to that internal call. It's very important to me to be able to memoize those calls, since they were the motivation of memoization in the first place: fullDetail calls are the most expensive calls (as seen in the aso module).

Also, this seems to make the tests fail.

icaroponce:
I'll test it again, idk why the tests failded, I had tested it sometimes before submit the pr.
Is it possible to know which test failed on Travis?

Since we are memoizing each method, I think that the app method called by parseList is memoized as well, so it'd cache each call of parseList too.

Let me check it once more and I'll let you know.

icaroponce:
@facundoolano , I've tested it a ton of times again. 
Some tests failed due to timeout at times, and other times tests didn't failed. 
(I know it's strange. but as it depends on my network connection and on goople play response time), 
Some test could take more than 5s to run on Travis.  

Maybe can we check if there was some timeout error when running the tests on travis?

About the `fullDetail: true` usage, if I got your point right, it's working as expected. 

Running the code above: 

```js
const gplay = require('../google-play-scraper').memoized();

gplay.list({
  category: gplay.category.GAME_ACTION,
  collection: gplay.collection.TOP_FREE,
  num: 1,
  fullDetail: true
})
  .then(items => items.map(item => item.offersIAP))
  .then(console.log);

gplay.list({
  category: gplay.category.GAME_ACTION,
  collection: gplay.collection.TOP_FREE,
  num: 1,
  fullDetail: true
});
```
I'm getting the following response:
```
DEBUG=google-play-scraper node gplay.js
  google-play-scraper Making request: {"url":"https://play.google.com/store/apps/category/GAME_ACTION/collection/topselling_free?hl=en&gl=us&num=1","method":"POST","form":{"start":0}} +0ms
  google-play-scraper Request finished +389ms
  google-play-scraper Making request: {"url":"https://play.google.com/store/apps/details?id=com.fortafygames.colorswitch&hl=en&gl=us"} +76ms
  google-play-scraper Request finished +423ms
[ true ]
[ true ]
```
As you can see, it didn't request the details or list again, it did only 2 requests, one for get the list, and the another one to get the apps detail, and, the next method call didn't make any requests, it did hit the cache.

facundoolano:
Yes, that example works, but I'm interested in other, subtler cases. Say you do: 

```js
gplay.search({term: "panda", fullDetail: true})
gplay.search({term: "panda game", fullDetail: true})
```

Both searches are likely to share most of the results, and thus the inner app detail calls will be the same, but since the `search` calls differ, no memoization will happen. The current implementation does cover this case. I've found many cases like this one while developing the aso module (not to say this scraper should accommodate to the other module's needs, but since we're supporting memoization, I prefer to take full advantage of it).

This would require to have some sort of initialization of the app function (or maybe the parseList one) and passing it around to the files that make use of it. I haven't thought it much myself yet, to be honest.

icaroponce:
Got your point, you're absolutely right. I haven't thought in this subtler case, and it'd be quite useful. 

I'll take a deeper look trough code and try to accomplish it. 

Feel free to close this PR in the meanwhile if you want.

icaroponce:
Hi, 

I took a deeper look through the code and couldn't find a reasonable solution for sharing the memoized modules internally, unfortunately. 

If anyone reading this pr have something in mind, please spread the word. 

I also wanna ask to @facundoolano what you think about merge this memoization setup  as it is, considering that we still have some opened issues and it is affecting directly the simple user trying to use and getting started to this module. And regarding to the compatibility to `aso` module, we don't need necessarily to update this module version there right now, we could do it as soon as we find a way to share the memoization internally.

facundoolano:
Closing this as a memoization refactor is already published.

Issue #160, opened 2017-07-21
feature: add an option to the search/list functions to include app permissions of each result


AdamSlack:
When setting ```fullDetail: true``` should the app's permissions not be included?

Example usage:
```
gplay.search({
        term: search_term,
        num: number_of_apps,
        throttle: per_second,
        fullDetail: true
    });
```

facundoolano:
Permissions are taken via a separate request than the application details, so including them would mean an extra request per app. I imagine duplicating the cost of the current `fullDetail` operations wouldn't be desirable in most cases, so to include permissions I'd rather add a separate option like `withPermissions` or something like that.

facundoolano:
I'll change the issue title to reflect that.

ormeno:
Hello , thanks for your code, it's great.

I write because the modification does not work me to get "list" with "permissions", I get the following error

_"   _throw new Error ('First argument to _arity must be a non-negative integer no greater than ten ');

_Error: First argument to _arity must be to non-negative integer no greater than tin
    at _arity (D: \ several \ Uni \ project \ node_modules \ google-play-scraper \ node_modu
les \ ramda \ dist \ ramda.js: 87: 19)
    at D: \ several \ Uni \ project \ node_modules \ google-play-scraper \ node_modules \ ramd
a \ dist \ ramda.js: 5003: 20
    at Object.f2 [as partial] (D: \ several \ Uni \ project \ node_modules \ google-play-s
craper \ node_modules \ ramda \ dist \ ramda.js: 473: 22)
    at Object. <anonymous> (D: \ several \ Uni \ project \ node_modules \ google-play-scrap
er \ index.js: 9: 21)
    at Module._compile (module.js: 413: 34)
    at Object.Module._extensions..js (module.js: 422: 10)
    at Module.load (module.js: 357: 32)
    at Function.Module._load (module.js: 314: 12)
    at Module.require (module.js: 367: 17)
    at require (internal / module.js: 16: 19)_"__

It's my first time with node js and I do not know how to solve it,  Can to be the changes start from a different version of the development?

Thanks so much

Rodrigo

facundoolano:
@ormeno please create a separate issue for this, and include all the required information.

ormeno:
Don`t worry, i've solved it. I had a problem of versions, i updated nodejs to 8.11.2 and version google-play-scraper to 3.1.1 and then it  works.

PR #161, opened 2017-07-23
Retrieve apps permissions when fullDetail is true on search method.

icaroponce:
Hi, this pr aims to fix the #160.

I've only added the implementation so far, if you agree with the changes proposed, let me know and I'll write tests and update the docs. 

Usage:

 ```js
gplay.search({
        term: 'panda',
        num: 2,
        fullDetail: true,
        withPermissions: true
    });
```
Added an option `permissionsShort` which might be used when the `withPermissions` is true.

### DIFF
--- a/lib/developer.js
+++ b/lib/developer.js
@@ -3,7 +3,7 @@
 const request = require('./utils/request');
 const memoize = require('./utils/memoize');
 const cheerio = require('cheerio');
-const getParseList = require('./utils/parseList');
+const {getParseList, getListWithPermissions} = require('./utils/parseList');
 
 // the pagination algorithm is very similar to search, so there's some -acceptable-
 // duplication. look out for chances to factor common stuff
@@ -78,6 +78,7 @@ function developer (opts) {
     }, opts);
 
     initialRequest(opts)
+      .then(getListWithPermissions(opts))
       .then(resolve)
       .catch(reject);
   });
--- a/lib/search.js
+++ b/lib/search.js
@@ -3,7 +3,7 @@
 const request = require('./utils/request');
 const memoize = require('./utils/memoize');
 const cheerio = require('cheerio');
-const getParseList = require('./utils/parseList');
+const {getParseList, getListWithPermissions} = require('./utils/parseList');
 
 function getClp (html) {
   // Try to find clp from "next page" html elem.
@@ -130,6 +130,7 @@ function search (opts) {
     };
 
     initialRequest(opts)
+      .then(getListWithPermissions(opts))
       .then(resolve)
       .catch(reject);
   });
--- a/lib/search.js
+++ b/lib/search.js
@@ -122,6 +122,8 @@ function search (opts) {
       country: opts.country || 'us',
       num: opts.num || 20,
       fullDetail: opts.fullDetail,
+      withPermissions: opts.withPermissions,
+      permissionsShort: opts.permissionsShort || false,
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: lib/utils/parseList.js [hunk 1]
```diff
@@ -27,7 +29,14 @@ function getParseDetailList (opts) {
         lang: opts.lang,
         country: opts.country,
         cache: opts.cache
-      });
+      })
+        .then((app) => {
```
facundoolano: this is coupled with the fullDetail option; note in line 9/11, if fullDetail is false the permissions options will be ignored

icaroponce: Yes, I think it should be coupled with `fullDetail` and only available if fullDetail is true, seeing that in this case the user is requesting even more details than what would be provided by fullDetail.

Do you think it should work regardless the `fullDetail` value?

icaroponce:
Thanks for reviewing it.

Yes, I'll add tests to it, as I've mentioned above.

lib/utils/parseList.js
facundoolano: Yes I think they should be independent, otherwise the API would be counterintuitive

icaroponce:
Updated this with another solution that works in all cases. 

Please, take a look at this, then I'll implement the tests and doc updates.

lib/developer.js
```diff
@@ -78,6 +78,7 @@ function developer (opts) {
     }, opts);
 
     initialRequest(opts)
+      .then(getListWithPermissions(opts))
```
facundoolano: add this one in the same place where parseList is

lib/search.js
```diff
@@ -122,12 +122,15 @@ function search (opts) {
...
       fullDetail: opts.fullDetail,
+      withPermissions: opts.withPermissions,
+      permissionsShort: opts.permissionsShort || false,
       price: opts.price ? getPriceGoogleValue(opts.price) : 0,
       throttle: opts.throttle,
       proxy: opts.proxy
     };
 
     initialRequest(opts)
+      .then(getListWithPermissions(opts))
```
facundoolano: same

lib/developer.js
icaroponce: I tried it firstly, but as you're controlling the `num` param in the `checkFinished` method, it'd request the permissions to all  the 49 ( or 48 ) apps, even if the number requested is lower (since the control is done by the `savedApps` array). I tried to change this behavior (adding the `splice` as it's being done in the `getDetailList` didn't worked) doing small changes in methods involved on this, but i couldn't.

facundoolano: Now that you mention it, I figure out that you shouldn't need to add a getListWithPermissions everywhere; the getParseList function should be smart enough to build a function that retrieves any combination of details and permissions according to what the options object specifies. This is not to say they should be coupled like in the first implementation, but that the resulting function should be built dynamically. That may require a bit of refactoring, but I think that's the best way to handle it.

icaroponce: Sure, I totally agree with you! I just wanted to be sure that we'd need to refactor the actual code to achieve it and I was not missing any obvious way to get it working. 

Since almost all methods are using this way to search the apps and controll the `savedApp` outside the `getParseList` and we'd need to modify them too, do you think is it worth to open a new base branch only for this refactoring?

facundoolano:
This got a bit outdated, feel free to send a new PR adapted to the latest code and what has been discussed here.

PR #164, opened 2017-07-31
Fixs cache param in search method when fullDetail is true

icaroponce:
Just realized that when the param `fullDetail` is `true` in `search` method, the `cache: false` param was not working as expected. This pr should fix it.


facundoolano:
thanks

PR #165, opened 2017-07-31
Sets num param default to 20 in developer method

icaroponce:
According to docs, the defalut value to `num` param in developer method should be `20`, however, it was `60`.


facundoolano:
Thanks, but I'd rather change the doc to match the code.

PR #167, opened 2017-09-01
Add developer physical address

koAlech:
Added parsing the developer's physical address in case it exists

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,7 @@ Results:
   developer: 'DxCo Games',
   developerEmail: 'dxcogames@gmail.com',
   developerWebsite: 'http://www.dxco-games.com/',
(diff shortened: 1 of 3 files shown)


facundoolano:
Thanks, can you add it to the README examples too?

koAlech:
I guess the only place that requires documentation is the results example for the app call. The problem is that the current example `com.dxco.pandavszombies` does not have a physical address therefore `developerAddress` is omitted in the response. 
What do you suggest to do? Add a different example with an address defined?

facundoolano:
Nah just use the address of a different developer in that same example.

koAlech:
Done

facundoolano:
Thanks!

koAlech:
😎

PR #181, opened 2018-01-19
Pass lang param to redirected URL when fetching similar apps

kblcuk:
Probably Google changed the way similar apps API works. So now request like
```
$ curl -I "https://play.google.com/store/apps/similar?id=com.chillingo.micromachines.android.gplay&hl=ru"
```
doesn't render HTML page with similar apps, but instead redirects like this:
```
HTTP/2 302
content-type: text/html; charset=UTF-8
...snip...
location: https://play.google.com/store/apps/collection/similar?clp=qgExCi8KKWNvbS5jaGlsbGluZ28ubWljcm9tYWNoaW5lcy5hbmRyb2lkLmdwbGF5EAEYAw%3D%3D:S:ANO1ljJ6vVU
```
The location it redirects to returns list of similar games, but using client's geo/language settings/etc.
Luckily that location also accepts `hl` parameter, so we can force which language we want back.

So essentially this PR
- allows to ask internal request library not to redirect
- appends `hl` param to location we redirect to, if such is found

I also added a simple test that tries to fetch similar apps using Russian, and then makes sure that at least some of the received summaries contain Cyrillic characters.

### DIFF
--- a/lib/similar.js
+++ b/lib/similar.js
@@ -13,11 +13,22 @@ function similar (getParseList, opts) {
     const appId = encodeURIComponent(opts.appId);
 
     const options = {
-      url: `https://play.google.com/store/apps/similar?id=${appId}&hl=${opts.lang}`,
-      proxy: opts.proxy
+      url: `https://play.google.com/store/apps/similar?id=${appId}`,
+      proxy: opts.proxy,
+      followRedirect: false // Don't follow redirects, because we need to add lang param to it
     };
 
     request(options, opts.throttle)
+      .then(response => {
+        if (response.statusCode > 300 && response.headers && response.headers.location) {
+          const redirectOpts = {
+            url: `${response.headers.location}&hl=${opts.lang}`,
+            proxy: opts.proxy
+          };
+          return request(redirectOpts, opts.throttle);
+        }
+        return response;
+      })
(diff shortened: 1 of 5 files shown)


facundoolano:
What happens if you omit the `clp=` parameter? I wonder if it's possible to use the right URL right away, instead of relying on the redirect

kblcuk:
Sadly no :(
```
$ curl -I https://play.google.com/store/collection/similar?id=com.chillingo.micromachines.android.gplay\&hl\=ru
HTTP/2 404
```

facundoolano:
The code looks good, but I'd like to dig around a bit more before merging. In the case of search the `clp` parameter is used for pagination, and it changes on each batch of results; I want to make sure we don't need that same logic here.

Also I'd like to make sure there's no way to bypass the redirect.

facundoolano:
I just did some tests on the browser, and as I expected, the new similar page is broken into multiple pages, using the same algorithm as the developer and search pages. So it's not enough to handle the first redirect, but we need to actually extract the new clp everytime and pass the proper parameters.

This algorithm is somewhat complex, and this will require some refactor so the pagination code is not duplicated across the three methods.

facundoolano:
Closing this, as it will be addressed as part of #188

Issue #182, opened 2018-01-23
A change in the site leads to incorrect behaviour


yura415:
Google recently changed how reCaptcha redirect works. Now it shows 302 Moved page with redirect to 503 page with recaptcha. So the scraper misunderstands this as an empty list and resolves with empty array.

Example contents of 302 Moved page:
```html
 <HTML><HEAD><meta http-equiv="content-type" content="text/html;charset=utf-8">
<TITLE>302 Moved</TITLE></HEAD><BODY>
<H1>302 Moved</H1>
The document has moved
<A HREF="https://ipv4.google.com/sorry/index?continue=https://play.google.com/store/apps/category/ANDROID_WEAR/collection/topselling_new_free%3Fhl%3Den%26gl%3Dal%26num%3D120&amp;hl=en&amp;q=EgQulH9UGIScndMFIhkA8aeDSxAT_ehABOGBG4kansp-C0dRCl3oMgFy">here</A>.
</BODY></HTML>
```

Solution is to put `followAllRedirects: true` in request options.

facundoolano:
Closing this since #183 was merged

PR #183, opened 2018-01-23
fix redirect to reCaptcha if response status is 302

yura415:
Fix for #182


facundoolano:
Thanks! Is there a test that fails for this? If not can you show an example code that would fail without this (or, if you have the time, a test that fails without your fix)?

yura415:
It doesn't breaks anything. Without this setting for example `list` request from IP that is banned on Play Store will resolve with empty array instead of throwing an error. The issue here is that actual code of the page will be 302 Moved redirect to ReCaptcha instead of 503 with that same ReCaptcha. So we need to follow that redirect in order to get correct error code from the page.

There is no test covering this. And maybe the only test case is to ban your IP by sending many requests to Play Store.

facundoolano:
Oh ok, got it now.

facundoolano:
Can you remove the package-lock?

facundoolano:
Thanks. I like the fact that you've added a requestOptions, in case the user needs to override some configuration. Can you remove the `proxy` options, though? since it's the kind of thing that should be handled through `requestOptions` from now on.

yura415:
Ok. I'll remove it.

yura415:
I've removed proxy option and also added requestOptions in developer & search initial requests.

Submitted PR for app-store-scraper with similar functionality: https://github.com/facundoolano/app-store-scraper/pull/44

facundoolano:
Thanks!

PR #187, opened 2018-02-07
fix requestOptions for search

yarafan:
Fix for passing requestOptions in search.js


facundoolano:
I don't think the code in `lib/utils/proxyOptions.js`, nor the socks dependency belongs in this project. Just expose a way to pass extra options to request, and clients of the library can use that to fill their proxy options.

Also this should apply to all methods that call request, not just search.

yarafan:
It was initially introduced several commits ago, so i rechecked it again and found bug only in search

facundoolano:
Ok, I see the problem now. requestOptions is not properly passed to the initialRequest in search.

facundoolano:
Fixed by eaff79d

Issue #192, opened 2018-03-15
TypeError: Cannot read property 'split' of undefined


figozhu:
version: 5.0.0

invode param:
```javascript
gPlay.list({
        category : gPlay.category.GAME,
        collection : gPlay.collection.NEW_FREE,
        lang : 'zh',
        country : 'CN',
        fullDetail : true,
        start : 0,
        num : 100,
    }).then(function (appList) {
        //console.log(appList);
        cb(null, appList);
    }).catch(function (err) {
        console.error("try to spider googleplay app list failed, err=", err);
        cb(err);
    });

```

log:
```
at parseFields (/googleplay/node_modules/google-play-scraper/lib/app.js:48:64)
    at tryCatcher (/googleplay/node_modules/bluebird/js/release/util.js:16:23)
    at Promise._settlePromiseFromHandler (/googleplay/node_modules/bluebird/js/release/promise.js:512:31)
    at Promise._settlePromise (/googleplay/node_modules/bluebird/js/release/promise.js:569:18)
    at Promise._settlePromise0 (/googleplay/node_modules/bluebird/js/release/promise.js:614:10)
    at Promise._settlePromises (/googleplay/node_modules/bluebird/js/release/promise.js:693:18)
    at Async._drainQueue (/googleplay/node_modules/bluebird/js/release/async.js:133:16)
    at Async._drainQueues (/googleplay/node_modules/bluebird/js/release/async.js:143:10)
    at Immediate.Async.drainQueues (/googleplay/node_modules/bluebird/js/release/async.js:17:14)
    at runCallback (timers.js:794:20)
    at tryOnImmediate (timers.js:752:5)
    at processImmediate [as _immediateCallback] (timers.js:729:5)
```

facundoolano:
Confirmed it fails.

Masterxv:
waiting for fix

abhijeet0007:
I got the same issue 

{
message: "Cannot read property 'split' of undefined"
}

timoweiss:
@facundoolano looks like google is doing some A/B testing. The structure and the css-class naming is different from time to time. Unfortunately, it looks like they are now generating css-class names which results in non-human-readable names (such as `AHFaub`). It could also be the case, that those names change everytime a change to the website is made :(

cgwyllie:
[...]

PR #193, opened 2018-03-15
Fixes TypeError: Cannot read property 'split' of undefined

roblav96:
Fixes a rare case of `TypeError: Cannot read property 'split' of undefined`

----

@facundoolano

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -42,6 +42,7 @@ function app (opts) {
 }
 
 function parseFields ($) {
+  if (!!$('head > base').attr('href')) return {};
   const detailsInfo = $('.details-info');
   const title = detailsInfo.find('.document-title').text().trim();
   const developer = detailsInfo.find('span[itemprop="name"]').text();


facundoolano:
Mmm I'm not convinced by this patch, it seems like it will just return an empty app object when the response is not what's expected.

We should look into why the response is some times different and try to avoid getting to that situation.

roblav96:
@facundoolano I agree. This was just a quick patch :P I'll have to look into why the selector returns `null`.

Masterxv:
{"message":"Cannot read property 'split' of undefined"}

fix it

Masterxv:
still not fixed

Masterxv:
waiting for fix

facundoolano:
As per discussion in #192 this is google using a different layout for some of the requests. I think a good provisional solution is to retry the requests (a limited number of times, and fail if it never comes back with the old layout) when the response has the new layout. 

Once they completely replace the old pages it will break again and we'll have to replace the app parsing code.

facundoolano:
Closing this as it's not a real fix to the underlying issue.

PR #206, opened 2018-04-05
clean html/xml tags from app description and added more missing fields

icaroponce:
There were some HMTL tags like `<br />` in the description text. This should clean them off.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -69,8 +69,9 @@ const MAPPINGS = {
   title: ['ds:3', 0, 0, 0],
   description: {
     path: ['ds:3', 0, 10, 0, 1],
-    fun: (text) => text.replace(/<[^>]*>/g, ''),
+    fun: (text) => text.replace(/<[^>]*>/g, '')
   },
+  descriptionHTML: ['ds:3', 0, 10, 0, 1],
   summary: ['ds:3', 0, 10, 1, 1],
   installs: ['ds:3', 0, 12, 9, 0],
   minInstalls: {
--- a/lib/app.js
+++ b/lib/app.js
@@ -67,7 +67,10 @@ const MAPPINGS = {
   // FIXME add appId
 
   title: ['ds:3', 0, 0, 0],
-  description: ['ds:3', 0, 10, 0, 1],
+  description: {
+    path: ['ds:3', 0, 10, 0, 1],
+    fun: (text) => text.replace(/<[^>]*>/g, ''),
+  },
   summary: ['ds:3', 0, 10, 1, 1],
   installs: ['ds:3', 0, 12, 9, 0],
   minInstalls: {
--- a/lib/app.js
+++ b/lib/app.js
@@ -70,7 +70,7 @@ const MAPPINGS = {
   title: ['ds:3', 0, 0, 0],
   description: {
     path: ['ds:3', 0, 10, 0, 1],
-    fun: (text) => descriptionText(text),
+    fun: descriptionText
   },
   descriptionHTML: ['ds:3', 0, 10, 0, 1],
   summary: ['ds:3', 0, 10, 1, 1],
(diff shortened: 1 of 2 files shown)


icaroponce:
Looks like only the next fields are missing now: 

```
maxInstalls
offersIAP
preregister
familyGenre 
familyGenreId
```

facundoolano:
Thanks, but I was thinking in a more proper html to string conversion for this field, maybe using some cheerio function. We'd want to convert <br> to new lines, decode special characters, etc.

facundoolano:
Thanks, but I was thinking in a more proper html to string conversion for this field, maybe using some cheerio function. We'd want to convert <br> to new lines, decode special characters, etc.

[this](https://github.com/icaroponce/google-play-scraper/blob/cb52bd086319d1851894dfdd8bd67cb4677d1e92/lib/app.js#L219-L223) could actually be used, by making a few changes to take a string instead of cheerio element.

(there may be more efficient/clean ways to achieve the same)

icaroponce:
I see, I'll check it then

lib/app.js
```diff
@@ -212,11 +217,11 @@ function extractFields (parsedData) {
 //   };
 
 // }
-// function descriptionText (description) {
-//   // preserve the line breaks when converting to text
-//   const html = '<div>' + description.html().replace(/<\/p>/g, '\n</p>') + '</div>';
-//   return cheerio.load(html)('div').text();
-// }
+function descriptionText (description) {
```
icaroponce: @facundoolano wdyt?

icaroponce:
Added `familyGenre` and `familyGenreId` props

lib/app.js
```diff
@@ -67,7 +68,11 @@ const MAPPINGS = {
   // FIXME add appId
 
   title: ['ds:3', 0, 0, 0],
-  description: ['ds:3', 0, 10, 0, 1],
+  description: {
+    path: ['ds:3', 0, 10, 0, 1],
+    fun: (text) => descriptionText(text),
```
facundoolano: you don't need a lambda here, can just reference `descriptionText`

icaroponce: You're right. Fixing in a moment.

lib/app.js
```diff
@@ -212,11 +219,11 @@ function extractFields (parsedData) {
...
 
 // }
-// function descriptionText (description) {
-//   // preserve the line breaks when converting to text
-//   const html = '<div>' + description.html().replace(/<\/p>/g, '\n</p>') + '</div>';
-//   return cheerio.load(html)('div').text();
-// }
+function descriptionText (description) {
+  // preserve the line breaks when converting to text
+  const html = cheerio.load('<div>' + description.replace(/\<br\>/g, '\r\n') + '</div>');
```
facundoolano: any particular reason to replace `br` instead of `p`? I honestly don't remember why it was p. I guess the important thing is that you make sure the output text is what we expect in terms of line breaks. (also I think sometimes the breaks are plain `<br>`. I'd say just pick a couple of example apps and see what's in their description.

lib/app.js
```diff
@@ -212,11 +219,11 @@ function extractFields (parsedData) {
...
+  return cheerio.text(html('div'));
```
facundoolano: I think `html.text()` should do here

lib/app.js
icaroponce: I also didn't understand why was `p` there... I tested with some apps, and got only `<br>`. I'll test with a detailed list to see if it varies.

lib/app.js
icaroponce: It doesn't

This is a example according to cheerio's doc:

```
const $ = cheerio.load('<div>This is <em>content</em>.</div>')
cheerio.text($('div'))
//=> This is content.
```

facundoolano: ok

lib/app.js
icaroponce: Did a quick test with 10 apps, all seems to be good. 

You can check: 

```
play.list({
  category: gplay.category.GAME_ACTION,
  collection: gplay.collection.TOP_FREE,
  num: 10,
  fullDetail: true,
})
  .then(data => data.forEach(x => console.log(x.description)))
```

icaroponce:
No idea why the tests still failing, locally they all pass.

facundoolano:
It can be some random failure, let me restart the build.

icaroponce:
Thanks 👍

PR #207, opened 2018-04-05
Added fields offersIAP and containsAds

icaroponce:
Added offersIAP field and also a field, which looks to be new, `containsAds`, whose purpose is to indicate whether the app has Ads or not.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -100,11 +100,11 @@ const MAPPINGS = {
   priceText: ['ds:8', 0, 2, 0, 0, 0, 1, 0, 2],
   offersIAP: {
     path: ['ds:3', 0, 12, 12, 0],
-    fun: (val) => !!val,
+    fun: (val) => !!val
   },
   containsAds: {
     path: ['ds:3', 0, 12, 14, 1, 0, 0, 1],
-    fun: (val) => !!val,
+    fun: (val) => !!val
   },
 
   size: ['ds:5', 0],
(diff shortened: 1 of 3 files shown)


lib/app.js
```diff
@@ -98,6 +98,14 @@ const MAPPINGS = {
   },
   currency: ['ds:8', 0, 2, 0, 0, 0, 1, 0, 1],
   priceText: ['ds:8', 0, 2, 0, 0, 0, 1, 0, 2],
+  offersIAP: {
+    path: ['ds:3', 0, 12, 12, 0],
+    fun: (val) => !!val
+  },
+  containsAds: {
```
facundoolano: I think this is the same as `adSupported` which I already added (although the paths differ I think conceptually it's the same). I think `containsAds` is a better name, but the previous version already uses `adSupported` so I think it's best to keep it.

lib/app.js
```diff
@@ -98,6 +98,14 @@ const MAPPINGS = {
   },
   currency: ['ds:8', 0, 2, 0, 0, 0, 1, 0, 1],
   priceText: ['ds:8', 0, 2, 0, 0, 0, 1, 0, 2],
+  offersIAP: {
+    path: ['ds:3', 0, 12, 12, 0],
+    fun: (val) => !!val
```
facundoolano: you can use `fun: Boolean` for this

lib/app.js
icaroponce: Ok, I've just removed the `containsAds` then

PR #208, opened 2018-04-06
Added preregister field and fixed some bugs

icaroponce:
Almost all bugs were happening when the app was a "`preregistered`" app, therefore there were some `undefined` values not properly handled. (When the `preregister` is `true`, the `price`, `comments` , `minInstall` and `histogram` values are undefined)

This `appId` can be used to test the cases described above: `com.gamedevltd.destinywarfare`

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -77,7 +77,7 @@ const MAPPINGS = {
   installs: ['ds:3', 0, 12, 9, 0],
   minInstalls: {
     path: ['ds:3', 0, 12, 9, 0],
-    fun: (text) => parseInt(text.replace(/[^\d]/g, ''))
+    fun: cleanInt
   },
   score: ['ds:10', 0, 0, 1],
   scoreText: ['ds:10', 0, 0, 0],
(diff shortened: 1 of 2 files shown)


lib/app.js
```diff
@@ -87,14 +87,18 @@ const MAPPINGS = {
     path: ['ds:10', 0, 1],
     fun: buildHistogram
   },
+  preregister: {
```
facundoolano: I'm not sure this is true. When you just upload a new app it doesn't have reviews, and thus won't have an histogram; that doesn't mean it's a preregister

facundoolano: I bet there's a boolean somewhere in there, although I admit it's going to be hard to track down. Is there somewhere we can find a list of 4 or 5 preregister apps so we can later test?

facundoolano: That doesn't seem to be the case, you can check these apps (with no downloads at the moment of writing, you can find others in the "top new paid" section):

https://play.google.com/store/apps/details?id=com.ThinkFast.CublastHD
https://play.google.com/store/apps/details?id=com.JacobDunbarGames.MayorMayorPremium

Those are flagged as preregister in your branch, although they are not.

icaroponce: Yes, it's true. I'm removing the `preregister` field for now, but I left the bugs fix there. I'll keep trying to track this.

icaroponce:
It looks like there's a random failure again on tests.

PR #221, opened 2018-04-26
Add privacy policy to app; Added coverage script with nyc

### LINKED ISSUE #171: Add Privacy policy to app
Retrive privacy policy link if exists when querying an app

srikanthlogic:
Fixes #171

### DIFF
--- a/test/lib.app.js
+++ b/test/lib.app.js
@@ -79,6 +79,13 @@ describe('App method', () => {
       });
   });
 
+  it('should get the privacy policy', () => {
+    return gplay.app({appId: 'com.snapchat.android'})
+      .then((app) => {
+        assert.equal(app.privacyPolicy, 'http://www.snapchat.com/privacy');
+      });
+  });
+
   it('should fetch app in spanish', () => {
     return gplay.app({appId: 'com.dxco.pandavszombies', lang: 'es', country: 'ar'})
       .then((app) => {
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ build/Release
 # Dependency directory
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
 node_modules
+.nyc_output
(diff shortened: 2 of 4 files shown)


test/lib.app.js
```diff
@@ -40,14 +40,14 @@ describe('App method', () => {
...
-        assert.equal(app.priceText, 'Free');
+        assert.equal(app.priceText, '');
         assert.equal(app.price, 0);
         assert(app.free === true);
         assert.equal(app.offersIAP, false);
         // assert(app.preregister === false);
 
         assert.equal(app.developer, 'DxCo Games');
-        assert.equal(app.developerId, 'DxCo+Games');
+        assert.equal(app.developerId, '8699533450056637394');
```
facundoolano: you changed this for the tests to pass, but it's actually a bug. Apparently the previous mapping doesn't work anymore. Please leave this test failing so we address it separately.

srikanthlogic: Thanks for the review. ,I remember seeing the ID; But shall cross check again and update PR in couple of days.

test/lib.list.js
```diff
@@ -89,8 +89,7 @@ describe('List method', () => {
       assert.isString(app.androidVersionText);
       assert.isString(app.androidVersion);
       assert.isString(app.contentRating);
-
-      assert.equal(app.priceText, 'Free');
+      assert.equal(app.priceText, '');
```
facundoolano: Again, this is failing because of a bug, if you don't fix the bug then don't hide the error.

srikanthlogic: On the HTTP response, I get the following 

[Resp.txt](https://github.com/facundoolano/google-play-scraper/files/1988037/Resp.txt)

Its empty and not "Free" -- [0,"USD",""].

facundoolano: Yes it's empty, but the code was supposed to be setting it to free. It's a bug we need to fix, so the test should keep failing until we do.

facundoolano: https://github.com/facundoolano/google-play-scraper/issues/230

Issue #225, opened 2018-04-27
Cannot read property 'replace' of undefined


sunnyrty:
"Cannot read property 'replace' of undefined"  new error

huoyinghui:
结构有变动，淡定

sunnyrty:
type in english

sunnyrty:
ok

sunnyrty:
ur change code works ?

EasyGitHub99:
@sunnyrty can you post some code here that produces this error? thanks

caucasianasian:
I have the same issue, using FullDetail: true when using gplay.list or gplay.developer. I think it's coming from the gplay.App function.

shellerbrand:
Unfortunately, some of the mappings on the Google Play Pages have changed. There is already a pull request https://github.com/facundoolano/google-play-scraper/pull/224 addressing the issue. 

If you are looking for an immediate fix, then this may be useful.

huoyinghui:
yee, we can see:
https://github.com/facundoolano/google-play-scraper/issues/223

facundoolano:
Please provide example code that produces the issue and the version of the module you are using.

victorfreitas:
@facundoolano See this image below, this error begins at line **178:51** of the file **lib/apps.js**, by which parameter **description** is arriving as undefined.

This error started this week. and there was no change in my application code.

**Node:** 8.11.1
**Express:** 4.16.3
**Google Play Scraper:** 6.0.1

![image](https://user-images.githubusercontent.com/21069978/39369133-dc8f961c-4a11-11e8-881d-60fa9e1e5aac.png)

sunnyrty:
@facundoolano did u fix it

sunnyrty:
Me fixed @huoyinghui  Code

droidmanspace:
Same problem again after fix. After fix it only works for 1h ,again same error

victorfreitas:
@droidmanspace Go back to the master version, it's working again.

droidmanspace:
ok thank you.

huoyinghui:
@facundoolano, ds:3 Has been restored 

'ds:3' : {id: 1.59150235E8 ,request:[["com.animocabrands.google.StarStableRun",7]
,2]
},'ds:4' : {id: 1.39809457E8 ,request:[null,["com.animocabrands.google.StarStableRun",7]

PR #226, opened 2018-04-27
Update app mappings to fix issue #225

victorfreitas:


huoyinghui:
@victorfreitas gp detail pages Has been restored at now, may be change in future

facundoolano:
Is this still relevant or not?

facundoolano:
Both master and this branch have tests failing, so I'm closing this. If anyone has the solution that actually fixes the problem, please send a separate PR (with the proper description).

PR #233, opened 2018-05-20
Added Privacy policy to app scrapper. Fixes #171

### LINKED ISSUE #171: Add Privacy policy to app

srikanthlogic:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "google-play-scraper",
-  "version": "6.0.1",
```
facundoolano: don't change the version as part of your PR, this should be done only when a version is actually published.

PR #237, opened 2018-06-02
Add option to specify language and country for suggest

filoozom:
PS: Had to bump `chai` to version 4 for `assert.notSameOrderedMembers`, but it did not break anything because the library is using `assert` and not `expect` or `should`.

### DIFF
--- a/lib/suggest.js
+++ b/lib/suggest.js
@@ -11,7 +11,7 @@ function suggest (opts) {
 
     const term = encodeURIComponent(opts.term);
     const options = Object.assign({
-      url: `https://market.android.com/suggest/SuggRequest?json=1&c=3&query=${term}`,
+      url: `https://market.android.com/suggest/SuggRequest?json=1&c=3&query=${term}&hl=${opts.lang || 'en'}&gl=${opts.country || 'us'}`,
       json: true,
       followAllRedirects: true
     }, opts.requestOptions);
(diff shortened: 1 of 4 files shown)


lib/suggest.js
```diff
@@ -11,7 +11,7 @@ function suggest (opts) {
 
     const term = encodeURIComponent(opts.term);
     const options = Object.assign({
-      url: `https://market.android.com/suggest/SuggRequest?json=1&c=3&query=${term}`,
+      url: `https://market.android.com/suggest/SuggRequest?json=1&c=3&query=${term}&hl=${opts.lang || 'en'}&gl=${opts.country || 'us'}`,
```
facundoolano: Can you do the `||` to set default options outside of the url? Check how it's done [here](https://github.com/facundoolano/google-play-scraper/blob/dev/lib/app.js#L16-L17).

facundoolano: Or better, using Object.assign (see [this](https://github.com/facundoolano/google-play-scraper/issues/238))

filoozom: Don't quite see where `Object.assign` would make sense, except maybe to completely copy `opts`, but I put it outside like for `term`.

filoozom:
Can you publish a new version for this too?

facundoolano:
published

PR #252, opened 2018-07-16
implement 'categories' method

### LINKED ISSUE #178: add a test to keep constants up to date
The test should request and scrape the Google Play website to obtain the current list of app categories and make sure they are the same as those in the constants file.

This way whenever they become outdated the tests will start to fail.

nonphp:
- extract list of categories from dropdown menu
- categories tests

Fixes #178

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,20 +5,19 @@ const constants = require('./lib/constants');
 const memoizee = require('memoizee');
 
 const appMethod = require('./lib/app');
-const categoriesMethod = require('./lib/categories');
 const getParseList = require('./lib/utils/parseList');
 const parseList = R.partial(getParseList, [appMethod]);
 
 const methods = {
   app: appMethod,
-  categories: categoriesMethod,
   list: R.partial(require('./lib/list'), [parseList]),
   search: R.partial(require('./lib/search'), [parseList]),
   suggest: require('./lib/suggest'),
   developer: R.partial(require('./lib/developer'), [parseList]),
   reviews: require('./lib/reviews'),
   similar: R.partial(require('./lib/similar'), [parseList]),
-  permissions: require('./lib/permissions')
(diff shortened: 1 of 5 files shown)


index.js
```diff
@@ -5,11 +5,13 @@ const constants = require('./lib/constants');
 const memoizee = require('memoizee');
 
 const appMethod = require('./lib/app');
+const categoriesMethod = require('./lib/categories');
 const getParseList = require('./lib/utils/parseList');
 const parseList = R.partial(getParseList, [appMethod]);
 
 const methods = {
   app: appMethod,
+  categories: categoriesMethod,
```
facundoolano: Put the require directly here instead of defining a variable, to follow the convention as the rest of the methods.

facundoolano: categories should also be defined in the memoized object below.

test/lib.categories.js
```diff
@@ -0,0 +1,21 @@
...
+describe('Categories method', () => {
+  it('should fetch valid list of categories', () => {
+    return gplay.categories().then(categoryIds => {
+      assert.isArray(categoryIds);
+      assert.isTrue(categoryIds.length > 0);
+    });
+  });
+
+  it('should have all categories from constant list of categories', () => {
+    return gplay.categories().then(categoryIds => {
```
facundoolano: It should also test the other way around: all categories should have their constant.

Suggestion: use [R.difference](https://ramdajs.com/docs/#difference) to do both checks with a set operation.

nonphp: Hi. What about categories that are not present in dropdown menu? -- APPLICATION and GAME
Maybe add them into resulting array as constants?

facundoolano: Yeah, that sounds reasonable

nonphp: Done

PR #259, opened 2018-08-06
Fix developerId parsing

philipbjorge:
Before this fix, the following was producing a 404 not found
```js
gplay.app({appId: 'com.dxco.pandavszombies'}).then(function(a) {
    gplay.developer({devId: a.developerId}).then(console.log);
});
```


facundoolano:
There's already a `developer` field which returns the developer name without encoding (i.e. a space instead of `+`). The name of the fields is admittedly confusing considering the use case you are showing, but I guess we need a rename rather than decoding the field.

facundoolano:
BTW it would also be useful to make a test with the code in your example

facundoolano:
Closing this due to inactivity.

PR #263, opened 2018-08-31
update mapping to fix error [score,scoreText,ratings,reviews,histogram]

lgc:


Hmachalani:
@facundoolano the lib is currently returning undefined for the score, scoreText etc.

lgc:
@facundoolano    I just update mappings,I don't know why test failures

lgc:
Hi @facundoolano 
very sorry for the late reply
From other pull requests, I know that the reason for the test failure is that ```com.dxco.pandavszombies``` has been deleted. This pull request resolves the ```score```, ```scoreText```, ```ratings```, ```reviews```, ```histogram``` fields without returning.

PR #265, opened 2018-09-01
add appId to permissioins response

taisuke-ohmi:
When executing the permissions method consecutively for multiple appid, I didn't know the result for which appid.
So I added appId to the permissions method.

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,14 +354,16 @@ gplay.permissions({appId: "com.dxco.pandavszombies"}).then(console.log);
 
 Results:
 ```javascript
-[ { permission: 'modify or delete the contents of your USB storage',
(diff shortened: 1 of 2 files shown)


taisuke-ohmi:
I rebase when https://github.com/facundoolano/google-play-scraper/pull/266 is merged. Then the test should pass.

facundoolano:
I don't get the need for this, you can extend the result with the app id without having to modify this method.

facundoolano:
```js
 gplay.permissions({appId: appId})
         .then((permissions) => ({permissions, appId}))
```

taisuke-ohmi:
I see, thanks.

PR #266, opened 2018-09-01
Fixed failed test case

taisuke-ohmi:
The app of "com.dxco.pandavszombies" has been deleted. Also, because there was an item that did not have a value, I fixed it.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -224,7 +224,7 @@ function extractComments (comments) {
   return R.compose(
     R.take(5),
     R.reject(R.isNil),
-    R.pluck(3))(comments);
+    R.pluck(4))(comments);
 }
 
 module.exports = app;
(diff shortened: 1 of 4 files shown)


facundoolano:
I see that you removed several checks, are those because of bugs in the mapings? Are those resolved in another PR?

facundoolano:
I don't like assertions commented, if the mappings changed, the tests are supposed to fail.

taisuke-ohmi:
Please try the app method once.
For example)
gplay.app({appId: 'jp.naver.line.android'})
  .then(console.log, console.log);
You will find that some items are "undefined". If you do not change the test you will need to fix bugs in mapping.

facundoolano:
That's ok, I rather have failing tests denoting something's broken. Please update the PR so no assertions are commented out.

taisuke-ohmi:
Since mapping was fixed, I rebased and returned the part commented out.
Also, since comments has not been acquired yet, I fixed it. Please also check https://github.com/facundoolano/google-play-scraper/pull/267

facundoolano:
cool, thanks!

PR #278, opened 2018-10-29
Default language fix

arkty:
Google play can return all reviews if hl option not set. No need to have default 'en' lang.

### DIFF
--- a/lib/reviews.js
+++ b/lib/reviews.js
@@ -9,18 +9,22 @@ const c = require('./constants');
 function reviews (opts) {
   return new Promise(function (resolve, reject) {
     validate(opts);
+    const form = {
+      pageNum: opts.page || 0,
+      id: opts.appId || opts.id,
+      reviewSortOrder: opts.sort || c.sort.NEWEST,
+      reviewType: 0,
+      xhr: 1
+    };
+
+    if(opts.lang) {
+      form.hl = opts.lang
+    }
 
     const options = Object.assign({
       method: 'POST',
       uri: 'https://play.google.com/store/getreviews',
-      form: {
-        pageNum: opts.page || 0,
-        id: opts.appId || opts.id,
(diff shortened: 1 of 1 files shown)


facundoolano:
This behavior is intentional, we want to have the same default language regardless of the place where the library is being run (otherwise different users will get different results running the same code).

facundoolano:
Not sure what that option would mean. We need to use something supported by the google play website.

arkty:
@facundoolano Google play supports requests without **hl** option. In this case, you can get all reviews in all languages.

I want to add option **all** into opts.lang (or remove default **en**, but it was rejected). 

There will be 3 cases:
1. opts.lang not set - default **en** language, as now.
2. opts.lang set to **all** - do not include **hl** into request (my case) and get all reviews.
3. opts.lang set to other value - use opts.lang in request, as now.

facundoolano:
Oh ok, I didn't realize that reviews behaved differently than the other pages

arkty:
@facundoolano so, should I add an **all** option or remove **default** hl when no lang specified (as now in PR)?

facundoolano:
`all` makes sense as an extra option. Make sure to mention it in the README.

yurganov:
@arkty, @facundoolano  you should be more careful with such behavior - not to set hl param at all. 

>>In this case, you can get all reviews in all languages.
If it so for now (and I am not sure), it can switch in any time to geo-based detection of IP.

arkty:
@yurganov hi! Thank you for research. 
I will try to reproduce it. Last time i got this result: different reviews when using 'ru' and w/0 hl. So, if it's now geo based, this should return same results.
I will double check it.

yurganov:
@arkty yes, i think different reviews between hl=ru and w/o hl is because of your IP.

And do not forget, that Google theoretically can think that your IP from another country.

facundoolano:
Closing as it looks like we can't rely on the default response sent by google.

PR #300, opened 2019-03-14
Add internal_developer_id to response

yurganov:
There is some apps, where developer_id is provided in developer URL "correctly":

ru.ok.android
image

But in some cases, it provided with encoded developer_name
air.com.bitrhymes.bingo
image

I've added new field to response with internal_developer_id, which is filled all time with numbers:

air.com.bitrhymes.bingo
image

ru.ok.android
image


yurganov:
Tests failed not because of me ;-)

Issue #299, opened 2019-03-08
node categories.js does not work


kani04:
* Operating System: Windows 10
* Node version: 20.15.1
* google-play-scraper version: 

### Description: 
node categories.js returns  [ 'APPLICATION' ] alone.
No furthur data is fetched

### Example code:
var gplay = require('google-play-scraper');

gplay.categories().then(console.log);

### Error message:

 [ 'APPLICATION' ]

PR #301, opened 2019-04-14
#299 fix categories list

amadi:


facundoolano:
Can you uncomment the related tests to confirm this works?

PR #302, opened 2019-04-24
[WIP] New search page

tanqhnguyen:
This is a work in progress. In our project, we notice that Google Play sometimes uses a new template for their search result page (They are doing A/B testing I guess). In the new template, they use the same initDataCallback approach as they are using now for the app page.

This PR tries to handle both cases. And I also refactor some code for better reusability

What is still missing
- Handle requests to fetch more results. The old clp approach doesn't work anymore
- Some more tests

I will try to finish this PR when I have some free time. Feel free to use this PR if you are having problem with your project. It only fetches the first 50 results, though

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -2,6 +2,7 @@
 
 const request = require('./utils/request');
 const matchScriptData = require('./utils/matchScriptData');
+const generateExtractFieldsFn = require('./utils/generateExtractFieldsFn');
 const queryString = require('querystring');
 const cheerio = require('cheerio');
 const R = require('ramda');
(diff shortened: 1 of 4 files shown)


tanqhnguyen:
> There are (a lot) of mappings in App, can't we use those? https://github.com/facundoolano/google-play-scraper/blob/d9e2d901bed3f572f7794b842123e5c75e36969c/lib/app.js#L67

Probably no because `ds:3` is now an array of items instead of a single object

facundoolano:
@tanqhnguyen can you pull from master and adapt this to use the new code added in #303 ?

tanqhnguyen:
This https://github.com/facundoolano/google-play-scraper/pull/307 might have covered everything here :) I will have to wait to see which one Google Play decides to use in the end (highly possible it's the new one)

PR #303, opened 2019-04-25
Make similar work

hunnapikes:
Google disable direct link to similar app page, but it can be obtained from the details request.


hunnapikes:
Similar method passes the test :)

lib/similar.js
```diff
@@ -1,28 +1,93 @@
...
+        }, opts.requestOptions));
+      })
+      .then(matchScriptData)
+      .then(extractFields)
+      .then(data =>
+        data.map(e => {
+          const q = url.parse(e.developerId, { parseQueryString: true });
+          e['developerId'] = q.query['id'];
+          e['url'] = new url.URL(e['url'], 'https://play.google.com').toString();
+          if (e['priceText'] === undefined) {
```
facundoolano: Can you handle all this processing declaratively as it's done in the app.js module? If fields require extra processing use mapping functions like this: 

https://github.com/facundoolano/google-play-scraper/blob/044f320cee8347481fd077543fe3d147a653c62f/lib/app.js#L79-L82

lib/similar.js
```diff
@@ -1,28 +1,98 @@
...
+  appId: [12, 0],
+  url: {
+    path: [9, 4, 2],
+    fun: (path) => new url.URL(path, 'https://play.google.com').toString()
+  },
+  icon: [1, 1, 0, 3, 2],
+  developer: [4, 0, 0, 0],
+  developerId: {
+    path: [4, 0, 0, 1, 4, 2],
+    fun: extaractDeveloperId
```
facundoolano:
```suggestion
    fun: extractDeveloperId
```

lib/utils/generateExtractFieldsFn.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+const R = require('ramda');
+
+module.exports = function generateExtractFieldsFn (mappings) {
```
facundoolano: Seems too small to keep in its own file. Whats more, it's related enough to the `matchScriptData` function that they could live both on the same module, and we should come up with a good name for that module.

hunnapikes: Any suggestions?

facundoolano: Maybe `scriptData` for the module name, so the code would look like

```js
request(options, opts.throttle)
    .then(scriptData.parse)
    .then(scriptData.extractor(MAPPINGS))
    // ...
```

(note I use `extractor` rather than `extract` because that function returns the actual function that does the extraction)

MojtabaRazmi: > Similar method passes the test :)

no similar method doesn't work yet!

PR #307, opened 2019-04-29
Fix search

hunnapikes:
Reuse similar request mappings.


facundoolano:
Can you resolve the conflicts?

lib/search.js
```diff
@@ -1,81 +1,63 @@
...
-  // extract the token for the next page request
-  // for the record, I hate regexps
-  // const s = html.match(/\\42(GAE.+?)\\42/);
-  const s = html.match(/\\x22-p6(.*?):S:(.*?)\\x22/g);
-  if (!s) {
-    return undefined;
-  }
-  return s[0].replace(/\\\\u003d/g, '=').replace(/\\x22/g, '');
-}
+const body = 'f.req=%5B%5B%5B%22qnKhOb%22%2C%22%5B%5Bnull%2C%5B%5B10%2C%5B10%2C50%5D%5D%2Ctrue%2Cnull%2C%5B96%2C27%2C4%2C8%2C57%2C30%2C110%2C79%2C11%2C16%2C49%2C1%2C3%2C9%2C12%2C104%2C55%2C56%2C51%2C10%2C34%2C77%5D%5D%2Cnull%2C%5C%22%token%%5C%22%5D%5D%22%2Cnull%2C%22generic%22%5D%5D%5D';
```
facundoolano: what is this exactly? can't it be extracted from the html instead of hardcoding it here?

lib/similar.js
```diff
@@ -2,9 +2,11 @@
 
 const request = require('./utils/request');
 const R = require('ramda');
-const url = require('url');
 const queryString = require('querystring');
 const scriptData = require('./utils/scriptData');
+const mapping = require('./utils/appListMapping');
+
+const process = R.partial(mapping.processList, [['ds:3', 0, 1, 0, 0, 0]]);
```
facundoolano: this API is a bit strange to read

lib/search.js
```diff
@@ -1,81 +1,63 @@
...
-  const $ = cheerio.load(html);
-  return opts.getParseList(opts)($)
-    .then((newApps) => savedApps.concat(newApps))
-    .then((apps) => checkFinished(opts, apps, nextToken, clp));
+function processAndRecur (html, opts, savedApps, rootPath) {
+  if (R.is(String, html)) {
+    html = scriptData.parse(html);
+  }
+
+  const apps = mapping.processList([...rootPath, 0], html);
```
facundoolano: I see that you only have two possible values for rootPath in the code below: `[0, 0]` and `['ds:3', 0, 1, 0, 0]`. Instead of passing the rootPath and then doing `...` can you use a more declarative style passing mapping objects?

e.g.

```js
{
    apps: [0, 0, 0],
    token: [0, 0, 7, 1],
}
```

If I understand correctly, by doing that you could also switch back to using the scriptData.extractor instead of this new processList

lib/search.js
```diff
@@ -1,81 +1,63 @@
...
-      gl: opts.country
-    },
-    // we need an unescaped qs for the requests to work
-    qsStringifyOptions: {encode: false},
-    followAllRedirects: true
+    body: body.replace('%token%', nextToken),
+    followAllRedirects: true,
+    headers: {
+      'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
+      'User-Agent': 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/74.0.3729.108 Safari/537.36'
```
facundoolano: is this agent necessary?

facundoolano:
What's the pagination situation with this? Does it work for all pages of search results?

I see there's one of the tests that still failing, is that a bug?

lib/search.js
hunnapikes: I don't know where it can be extracted. If you find I can fix it.

lib/search.js
hunnapikes: No.

hunnapikes:
> What's the pagination situation with this? Does it work for all pages of search results?
> 
> I see there's one of the tests that still failing, is that a bug?

Test case 6 sometimes failing, maybe not unique apps.

lib/utils/appList.js
```diff
@@ -0,0 +1,54 @@
...
+function extractor (root) {
+  return R.partial(extract, [root]);
+}
+
+function extract (root, data) {
+  const input = R.path(root, data);
+  return R.map(scriptData.extractor(MAPPINGS), input);
+}
+
+module.exports = Object.assign({MAPPINGS, extractor, extract});
```
facundoolano: No need to Object.assign here I think, exporting `{MAPPINGS, extractor, extract}` should be the same

lib/similar.js
```diff
@@ -43,55 +44,14 @@ function similar (opts) {
...
-        const input = R.path(['ds:3', 0, 1, 0, 0, 0], data);
-        return R.map(scriptData.extractor(MAPPINGS), input);
-      })
+      .then(extract)
       .then(resolve)
       .catch(reject);
   });
 }
 
-const PATH_MAPPING = {
```
facundoolano: PATH_MAPPING is a more clear name considering the rest of the codebase

lib/utils/appList.js
facundoolano: Also, it doesn't seem like you ever use MAPPINGS outside of this module, so it shouldn't be exported.

lib/utils/appList.js
```diff
@@ -0,0 +1,54 @@
...
+  scoreText: [6, 0, 2, 1, 0],
+  score: [6, 0, 2, 1, 1]
+};
+
+function extaractDeveloperId (link) {
+  const q = url.parse(link, {parseQueryString: true});
+  return q.query['id'];
+}
+
+function extractor (root) {
```
facundoolano: I find a bit confusing that you have both extractor _and_ extract and both are exported. I would prefer to have just one of them exported, and with a short comment explaining what it does.

facundoolano:
thanks!

PR #315, opened 2019-05-30
es5 support

pedropapa:
Allows old browser clients which don't have ES6 support to run the lib. It's most useful for mobile apps, which may run in old devices that still have an old WebView version (Like in Android 4.4 ~ 5.1 devices). All I did was to install babel dev dependency and created a build command which generates/updates an `es5` directory with the transpiled ES6 code from `lib` directory.


facundoolano:
Do you maybe want to add a description to this? It's a big PR and I can't tell what it does.

pedropapa:
Sorry, done.

facundoolano:
There are a number of things that don't convince me about this. 

First if the code is autogenerated by a command (and should be refreshed when the code changes), I don't think that code should be committed into the repository.

Second, running in the browser is not a primary concern of this project. I'm hesitant to extend its scope by adding more code/configuration/dependencies. You could have a separate google-play-scraper-browser or whatever that includes this and extends as necessary to run in the browser.

Lastly, if porting to the browser is something that can be easily achieved by using another library and some configuration, I would prefer to just add a section in the readme with instructions on how to do it, for those that are interested.

PR #318, opened 2019-06-27
Adds permission mapper

bluetrickpt:
This PR adds a mapping between permission values and the permission "names". 

The mapping is found in `permissionMapper.js` and currently only supports the 'en' language. Details on how the mapping was made (and therefore how it can be made for other languages) is found in [https://github.com/bluetrickpt/permission-description2name](https://github.com/bluetrickpt/permission-description2name). Translating is easier since the mapping itself is already made. One just have to change the keys of the  object.

It is possible that the mapping does not have all existing permissions. This is because old apps in gplay will have permissions that no longer exist. However, adding entries to the mapping is quite easy, one just needs to update `permissionMapper.js`.

**Note:** since I've only created the mapping in 'en', when using a different language, the field `permissionValue`will be empty for all returned permissions.

Let me know if you want any change. For example, we could instead of sending the field `permissionValues` with empty values for non translated languages, not send this field at all in the object. I can also add other languages as needed as I think that will be quite fast now that I have everything set up.

Thanks!

### DIFF
--- a/lib/permissions.js
+++ b/lib/permissions.js
@@ -1,20 +1,22 @@
 'use strict';
 
 const request = require('./utils/request');
+const permissionName2ValueMapper = require('./utils/permissionMapper');
 const R = require('ramda');
 
 const isPermission = (item) => R.is(Array, item) && item.length === 3 && R.is(String, item[0]) && R.is(String, item[1]);
 
-function parsePermissions (items) {
+function parsePermissions (items, lang) {
   return items.reduce((results, item) => {
     if (isPermission(item)) {
       return R.append({
         permission: item[0],
-        description: item[1]
+        description: item[1],
+        permissionValue: permissionName2ValueMapper(item[0], lang),
       }, results);
     }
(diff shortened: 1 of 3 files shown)


facundoolano:
Thanks for this work. I have to say I'm a bit hesitant about including this in the library, I'm not entirely sure it fits here. Why do you consider that this mapping belongs here versus having a separate library that calls this one and does the mapping?

bluetrickpt:
Sorry for the delay. Was on a short holidays.
I agree with you. I wanted to extract the names of the permissions as defined in Android developers without downloading the APKs and that's why this idea came to mind. I also saw some discussion of people trying to do the same but ended up having to download the APKs (unfortunately I can't seem to find this discussion). At the same time, I ended up needing to download the APKs to extract other info, so I'm also not using this.
Nevertheless, making this a library might be a good idea. I'll do that when I have some free time. 

Thanks and sorry for the inconvenience.

Issue #284, opened 2018-12-14
developers api is not working


farzan3:
api is not working with developer name .. i think again google changed template .
i check parseList.js and i am sure template is changed . 
I am c# developer i wish i can help with nodejs :)

Thanks for any kind of help 

* Operating System:
* Node version:
* google-play-scraper version:

### Description: 
Describe your issue here

### Example code:
```js
// Put code to reproduce the issue here
```

### Error message:

```
Put error message here
```

vitaliy901:
For me not working too :(

vitaliy901:
Maybe someone gotta write a fix ? Please guys

Felixaep:
Another one that does not work

farzan3:
please fix it

facundoolano:
uhm I don't know folks, maybe if you send another ten comments here and keep asking in twitter I'll pause the rest of my life and fix your issue.

vitaliy901:
> uhm I don't know folks, maybe if you send another ten comments here and keep asking in twitter I'll pause the rest of my life and fix your issue.

sorry, bro )

farzan3:
hey guys i can fix it :D but that will not be professional 🗡

vitaliy901:
> hey guys i can fix it :D but that will not be professional 🗡

Ok, send it in Pull Request, but keep Object keys the same )

vitaliy901:
> hey guys i can fix it :D but that will not be professional 🗡

it will be very helpfull )

farzan3:
> > hey guys i can fix it :D but that will not be professional 🗡
> 
> it will be very helpfull )

just did that . its working for me :) i am c# dev not nodejs 👯‍♂️

vitaliy901:
> done dude

thanks )

dsharkov:
Guys, do you have any updates on this issue?
Have the same issue - call "gplay.developer({devId: "Netmarble"}).then(console.log);" and get an empty collection((

Felixaep:
Hi, I tryed to fixed this ... But i need change de devId by developerId and now this parameter is a number:
example: https://play.google.com/store/apps/dev?id=6577204690045492686 

this is the changes for developer.js

`'use strict';

const request = require('./utils/request');
const cheerio = require('cheerio');

function getClp (html) {
  // Try to find clp from "next page" html elem.
  let match = html.match(/\?clp=(.*?)">/);
  // ... if we don't have it, we're probably on innerPage;
[...]

PR #322, opened 2019-08-11
Fix developers API

Suvitruf:
https://github.com/facundoolano/google-play-scraper/issues/284#issuecomment-520255165

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,6 @@ build/Release
 # https://www.npmjs.org/doc/misc/npm-faq.html#should-i-check-my-node_modules-folder-into-git
 node_modules
 .nyc_output
+
+# PhpStorm/WebStorm config
+.idea
--- a/lib/developer.js
+++ b/lib/developer.js
@@ -100,8 +100,7 @@ function buildInitialUrlInt (opts) {
 }
 
 function buildInitialUrlStr (opts) {
-  const devId = encodeURIComponent(opts.devId);
-  return `https://play.google.com/store/apps/developer?id=${devId}&hl=${opts.lang}&gl=${opts.country}&term=${devId}`;
+  return `https://play.google.com/store/apps/developer?id=${opts.devId}&hl=${opts.lang}&gl=${opts.country}`;
 }
 
 function initialRequestStr (opts) {
--- a/lib/developer.js
+++ b/lib/developer.js
@@ -1,44 +1,57 @@
 'use strict';
 
 const request = require('./utils/request');
-const cheerio = require('cheerio');
+const R = require('ramda');
+const scriptData = require('./utils/scriptData');
+const appList = require('./utils/appList');
 
-// the pagination algorithm is very similar to search, so there's some -acceptable-
-// duplication. look out for chances to factor common stuff
-function getNextToken (html) {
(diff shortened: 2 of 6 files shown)


Suvitruf:
@facundoolano, even if it's working, probably it doesn't feet the rest project code style, etc. Would be nice to see any recommendations/suggestions (:

facundoolano:
So is it working? Can you summarize what the implementation is now? I see that you had to increase the mocha timeout, did the method get much slower now?

Suvitruf:
> 
> 
> So is it working? 

Yeh, I use this implementation right now for our project, and it works. 

> Can you summarize what the implementation is now?

1. We load developer page. It could be in different format:

   - With string id, e.g. https://play.google.com/store/apps/developer?id=Google%20LLC
   -  Int id, e.g. https://play.google.com/store/apps/dev?id=5700313618786177705

2. Get [cluster link](https://github.com/facundoolano/google-play-scraper/blob/dev/lib/search.js#L61) from button, e.g. `https://play.google.com/store/apps/collection/cluster?clp`.
3. Send requests like in [`search` ](https://github.com/facundoolano/google-play-scraper/blob/dev/lib/search.js).
And receive response, but it is not html, and not fully valid json (there are some strange symbols in the beginning of the response). We [cut](https://github.com/facundoolano/google-play-scraper/blob/dev/lib/search.js#L47) them, and then parse it as json and [parse apps using `REQUEST_MAPPINGS`](https://github.com/facundoolano/google-play-scraper/blob/dev/lib/search.js#L49).

> I see that you had to increase the mocha timeout, did the method get much slower now?

I've increased it, 'cause i passed `fullDetail: true`

Suvitruf:
> did the method get much slower now?

I haven't tested it. I mean, for me, there were no choice, 'cause there is no working alternative. I can check it later to see, if I can optimize some things.

xenio:
It is working for me. I get all my apps as developerID

facundoolano:
I'm inclined to merge this now, but there's one of the developer tests that is still failing, the one that checks that multiple pages of results are fetch, can you fix that one too?
By the looks of the website, there are less than 100 apps from Google LLC, so I guess it makes sense that it fails if we ask for >= than 100

Suvitruf:
Sure, I'll going to check it on my day off.

facundoolano:
Looks like two method fail now, I wonder if the recent changes in the website have broken this as well?

icarcal:
Hey, guys.
Google seems to have two different urls:
- The first one:
  - https://play.google.com/store/apps/dev?id=7908612043055486674&hl=en (probably for some opt-in setting)
- The second one
  - https://play.google.com/store/apps/developer?id=WhatsApp+Inc.&hl=en

I think that we need to add a fallback to this for now, for example, search for the new one with the devId, if the URL is not found, check for the second URL.

The only problem with this approach is that both methods can fail, leaving you with no information about the developer apps.

Unfortunatelly this seems like another breaking change :disappointed:

I'll try to come up with something for this case too.

icarcal:
@facundoolano @Suvitruf check the PR #344 
Finally the building is passing :tada:

facundoolano:
fixed by #344

PR #323, opened 2019-08-15
Update MAPPINGS ds:7 change to ds:6

lgc:


facundoolano:
Are none of these fields affected by the tests?

lgc:
> Are none of these fields affected by the tests?

Google Play has just changed

PR #328, opened 2019-08-19
fix developer id

alkattan:
Developer id is returning as Khan Academy for the search term khan instead of Khan+Academy
This pull request fixes the developer id parsing

### DIFF
--- a/lib/utils/appList.js
+++ b/lib/utils/appList.js
@@ -32,7 +32,7 @@ const MAPPINGS = {
 };
 
 function extaractDeveloperId (link) {
-  return querystring.parse(link.split('?')[1]).id;
+  return link.split('?id=')[1];
 }
 
 /*


lib/utils/appList.js
```diff
@@ -32,7 +32,7 @@ const MAPPINGS = {
 };
 
 function extaractDeveloperId (link) {
```
facundoolano: while you are at it can you 
```suggestion
function extractDeveloperId (link) {
```

facundoolano: (and update the calling sites)

alkattan: Sorry, I didn't get what you mean.

facundoolano: there's a typo in the function name `extAract` instead of `extract`

Issue #336, opened 2019-09-08
Top apps error: URL has changed, or doesn't exist anymore?


knife-dev:
* Operating System: MacOS
* Node version: v10.12.0
* google-play-scraper version: 6.2.8

### Description: 
Empty list, nothing returned at all.

### Example code:
```js
gplay.list({ fullDetail: true,
  category: 'APPLICATION',
  country: 'ar',
  num_start: 0,
  start: 0,
  collection: 'topselling_free',
  lang: 'en');

```

### Error message:
- None, empty list, (I did some debugging and https://play.google.com/store/apps/category/ART_AND_DESIGN/collection/topselling_free?hl=en&gl=ar&num=9 doesn't exist anymore)
```
Error 404 page not found
```

facundoolano:
duplicate of #333

Issue #334, opened 2019-09-06
Error: App not found (404) For List api 


andylauxugang2:
* Operating System:Ubuntu lastest
* Node version:v10.16.3
* google-play-scraper version:

### Description: 
When I use the List api to test your official demo. That show me the error below:

### Example code:
```js
var gplay = require('google-play-scraper');

gplay.list({
    category: "FAMILY",
    num: 2
  })
  .then(console.log, console.log);

```

### Error message:

```
{ Error: App not found (404)
    at /home/sysop/nodejs/node_modules/google-play-scraper/lib/utils/request.js:44:19
    at process._tickCallback (internal/process/next_tick.js:68:7) status: 404 }
```

dipbhi:
I am getting same error for using the given example in the help doc.

`
var gplay = require('google-play-scraper');

gplay.list({
    category: gplay.category.GAME_ACTION,
    collection: gplay.collection.TOP_FREE,
    num: 2
  })
  .then(console.log, console.log);
`

 It's likely that Google Play has changed the list page format.

My environment:
- OS: Windows 10 Pro
- Node.js: v10.15.0 
- google-play-scraper: 6.2.8

vipinhelloindia:
I am getting the same error for 
var gplay = require('google-play-scraper');

gplay.reviews({appId: 'com.jeevansathi.android'})
  .then(console.log, console.log);

facundoolano:
duplicate #333

Issue #333, opened 2019-09-04
suddenly fires error 'App not found (404)'  on request


gamechosun:
* Operating System: Windows7, centos
* Node version: 10.15.3
* google-play-scraper version: 6.2.7

Description: 
I implemented this module in my service and worked well. 
recently, module fires 'App not found(404)' error 70~80% of request. 
strange thing is request success sometime.

Example code:
var G = require('google-play-scraper');
G.list({
    category: G.category.GAME,
    collection: G.collection.GROSSING,
    start: 0,
    num: 20,
    lang: 'ko',
    country: 'kr',
    throttle: 10
}).then(console.log, console.log)

Error message:

{ Error: App not found (404)
    at d:\workspace\googleplayscraper\node_modules\google-play-scraper\lib\utils\request.js:44:19
    at process._tickCallback (internal/process/next_tick.js:68:7) status: 404 }

vitaliy901:
yes have that error on category only page too ... :(

marxsk:
I have the very same issue when I attempt to download reviews.

takeTrace:
same issue + 1

xulifan:
I think Google updated the links of play store today.

mitchross:
Bump on this too

marxsk:
Can anyone confirm that updated version should use /batchexecute (used by web browser) ?

vitaliy901:
@xulifan i see that links not changes ( but it doesn't works... something strange. Request is throw 404, but when i try curl on my ubuntu server it's opening (

marxsk:
@xulifan can you pls send me your curl command? So I can take a closer look at it?

xulifan:
@marxsk For example, for collection TOP GROSS and category GAME, the generated URL https://play.google.com/store/apps/category/GAME/collection/topgrossing?hl=en&gl=jp is no longer available. 
From https://play.google.com/store/apps/top/category/GAME , we can see the link for TOP GROSS game is https://play.google.com/store/apps/collection/cluster?clp=0g4YChYKEHRvcGdyb3NzaW5nX0dBTUUQBxgD:S:ANO1ljLhYwQ&gsr=ChvSDhgKFgoQdG9wZ3Jvc3NpbmdfR0FNRRAHGAM%3D:S:ANO1ljIKta8

andylauxugang2:
Bump on this too for me

icarcal:
Hey guys!
Had the same problem.
[...]

PR #337, opened 2019-09-09
fixes for gplay store after 2019-09-01 updates

icarcal:
This PR fixes the following issues: 
#333 
#334 
#335 
#336 

I have updated the following methods:
 ```.list```
 ```.reviews```
 ```.permissions```
 ```.categories```

 ```.list ``` and  ```.reviews ``` methods introduce some breaking changes, since google allow you to fetch data only through ```batchexecute``` endpoint.
You can now only send the ```num``` parameter (it will allow you to fetch the desired number of category apps and reviews).
The ```.list``` method is unstable right now, since google sometimes respond with 200 apps, and sometimes more (for example 600 and so on).

For the ```.permissions``` method I've added the ```type``` parameter to the response (since this is not a breaking change but add value to it).
For the ```.reviews``` method I've added the ```thumbsUp``` and ```version``` parameters, this is also not a breaking change.

I've also updated the tests and docs

PR #339, opened 2019-09-11
fix .list method (store update 2019-09-01)

icarcal:
#333
#334
#336

```.list```

.list method introduce some breaking changes, since google allow you to fetch data only through ```batchexecute``` endpoint.
You can now only send the ```num``` parameter (it will allow you to fetch the desired number of category apps).
The ```.list``` method is unstable right now, since google sometimes respond with 200 apps, and sometimes more (for example 600 and so on).

Check #333 for more info about this


vitaliy901:
I have one issue, gp.list({collection: 'topselling_free', num: 100}).then(res => {
    console.log(res);
})

Gives me Error: 
 TypeError: Cannot read property 'topselling_free' of undefined

icarcal:
@vitaliy901 thanks!
Gonna to look into it!

icarcal:
@vitaliy901 fixed ```.list``` for top collections.
Can you check if this is working correctly right now?

vitaliy901:
@irarcal still same - gp.list({collection: gp.collection.TRENDING, num: 100}).then(res => {
    console.log(res);
})

TypeError: Cannot read property 'movers_shakers' of undefined

i see that it doesn't work, when there is no opts.category ...

vitaliy901:
@icarcal

vitaliy901:
@icarcal parameter start, now doen't work too ?

icarcal:
@vitaliy901 no.
This is a breaking change,
You can now only send the `num` parameter (it will allow you to fetch the desired number of category apps).

You can check a little bit more about this discussion at #337

vitaliy901:
@icarcal i think for working test and another api package "google-play-api", start option should work

Issue #335, opened 2019-09-06
Error "App Not Found (404) " for gplay.permissions and gplay.reviews


abubakaraziz:
* Operating System:Ubuntu 18.04.2 LTS
* Node version:v10.15.3
* google-play-scraper version:6.2.8

### Description: 
I was using Gplay. reviews and Gplay.permissions and recently both of them are giving "App not found(404)" on all the requests.

### Example code for permissions:
var gplay = require('google-play-scraper');
 gplay.permissions({appId: 'word.cross.game.wordcrossy.free'})
 .then(console.log);
### Error message:

Error: App not found (404)  at /home/aziz/Network_Security_Project/googletest/node_modules/google-play-scraper/lib/utils/request.js:44:19 at process._tickCallback (internal/process/next_tick.js:68:7) // for permissions

### Example code for Reviews:
var gplay = require('google-play-scraper');
   gplay.reviews({
    appId: 'word.cross.game.wordcrossy.free',
    page: 0,
    sort: gplay.sort.RATING
  }).then(console.log, console.log);
~                                        
### Error message:
{ Error: App not found (404)
    at /home/aziz/Network_Security_Project/googletest/node_modules/google-play-scraper/lib/utils/request.js:44:19
    at process._tickCallback (internal/process/next_tick.js:68:7) status: 404 }

MojtabaRazmi:
yes i have this problem also

chekanskiy:
yes, i'm getting the same error

PR #340, opened 2019-09-11
fix .permissions method (store update 2019-09-01)

icarcal:
#335 

```.permissions```
 ```.categories```

I've added the ```type``` parameter to the response (since this is not a breaking change but adds value to it).

### DIFF
--- a/lib/permissions.js
+++ b/lib/permissions.js
@@ -1,24 +1,6 @@
 'use strict';
 
-const request = require('./utils/request');
-const R = require('ramda');
-
-const isPermission = (item) => R.is(Array, item) && item.length === 3 && R.is(String, item[0]) && R.is(String, item[1]);
-
-function parsePermissions (items) {
-  return items.reduce((results, item) => {
-    if (isPermission(item)) {
-      return R.append({
-        permission: item[0],
-        description: item[1]
-      }, results);
-    }
-    if (R.is(Array, item)) {
-      return results.concat(parsePermissions(item));
-    }
-    return results;
-  }, []);
-}
+const { processPermissions } = require('./requesters/permissionMappedRequests');
 
 function permissions (opts) {
   return new Promise(function (resolve, reject) {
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: lib/utils/permissionList.js [hunk 1]
```diff
@@ -0,0 +1,25 @@
+'use strict';
+
+const R = require('ramda');
+const scriptData = require('./scriptData');
+
+function getPermissionMappings (type) {
+  const MAPPINGS = {
+    permission: [1],
+    description: [1],
```
facundoolano: if both have the same value, why are they separate fields?

lib/permissions.js
```diff
@@ -1,60 +1,18 @@
...
-        description: item[1]
-      }, results);
-    }
-    if (R.is(Array, item)) {
-      return results.concat(parsePermissions(item));
-    }
-    return results;
-  }, []);
-}
+const { processPermissions } = require('./requesters/permissionMappedRequests');
```
facundoolano: Why is most of the logic in this separate file, if this one is almost empty?

lib/utils/permissionList.js
icarcal: @facundoolano 
Google does not provide "permission" field anymore, only description.
So, for backward compatibility I used the same field for both permission and description.
I can comment on this too.

You can check the discussion and response return [here](https://github.com/facundoolano/google-play-scraper/pull/337#issuecomment-529534740) 

May I suggest a file of changeslog?
Where we can add what was changed between versions, instead of comment each file?

lib/permissions.js
icarcal: @facundoolano 
I tried to follow a "pattern", leaving only the initial request in the base files.
So in the end is basically
 - root files for methods: initial request + validation (what can be separated as well)
 - requests in the /requesters folder
 - parsers in the /utils folder (the can be changed to parsers folder, for better "readability")

lib/utils/permissionList.js
facundoolano: A changelog would certainly be nice, I just never got around to it. 

I don't think we need to keep things around for backward compatibility. Being a scraper we have to break things all the time anyway to keep up with the page changes.

PR #341, opened 2019-09-11
fix .reviews method (store update 2019-09-01) 

icarcal:
#335 

 ```.reviews```
 
 ```.reviews ``` method introduce some breaking changes, since google allow you to fetch data only through ```batchexecute``` endpoint.
You can now only send the ```num``` parameter (it will allow you to fetch the desired number of reviews).

For the ```.reviews``` method I've added the ```thumbsUp```, ```version``` and ```criterias``` parameters, this is not a breaking change.


lib/mappers/request.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
```
facundoolano: Are these "global" mappings (i.e. mappings that apply to more than one specific method?).

Can you add a heading comments explaining what are these for (how they are different from the method-specific mappings).

Also the word "request" is kind of generic and is used with other meaning elsewhere in the codebase, so we should come up with a differnt one.

lib/requesters/reviewsMappedRequests.js
```diff
@@ -0,0 +1,91 @@
...
+
+const debug = require('debug')('google-play-scraper:reviewsMappedRequests');
+const R = require('ramda');
+const request = require('../utils/request');
+const scriptData = require('../utils/scriptData');
+const reviewsList = require('../utils/reviewsList');
+const { REQUEST_MAPPINGS } = require('../mappers/request');
+const { BASE_URL } = require('../utils/configurations');
+
+const TOKEN_TYPES = {
```
facundoolano: a comment here would help

lib/requesters/reviewsMappedRequests.js
```diff
@@ -0,0 +1,91 @@
...
+  initial: 'initial',
+  request: 'request'
```
facundoolano: again, request may not be the best name

lib/requesters/reviewsMappedRequests.js
```diff
@@ -0,0 +1,91 @@
...
+};
+
+function getBodyForRequests ({
+  appId,
+  sort,
+  numberOfReviews = 100,
+  withToken = '%token%',
+  tokenType = 'initial'
```
facundoolano: shouldn't this be `TOKEN_TYPE.initial`?

lib/requesters/reviewsMappedRequests.js
```diff
@@ -0,0 +1,91 @@
...
+}) {
+  const formBody = {
```
facundoolano: comment explaining what's this used for

facundoolano:
@icarcal we have `num` now, but don't we need also a `start` as in other methods? how do I get, say, the second 100 reviews? Do I need to fetch 200? That doesn't seem to scale well

icarcal:
@facundoolano every time you'll have to fetch the 200 reviews, even with the `start` param.

I double checked it on google play, and the requests have the following body ('till page 3):
```
[[["UsvDTd","[null,null,[2,2,[40,null,null],null,[]],[\"com.instagram.android\",7]]",null,"generic"]]]
```
```
[[["UsvDTd","[null,null,[2,2,[40,null,\"CoABCn4qbQo59-ynvbf_2yVsmJ6WnsXPz8_Pz8_Py8ebmZrKxsuaxc7FiZLFyszPzsnKz83LzMfNys3NyMzJyP_-EFAhbf9dDYzEU6sxJMzJNyuJ8x45k9okAEhCWBNQAFoLCR-mVLb9M1KtEANg_bz5wQQyDQoLCgAo8sKV15Hd5AI\"],null,[]],[\"com.instagram.android\",7]]",null,"generic"]]]
```
```
[[["UsvDTd","[null,null,[2,2,[40,null,\"CoABCn4qbQo59-ynvyr_yoqKmJ6WnsXPz8_Pz8-emcuZzpvMmcuZxc7FiZLFyszPzsnKz83LzMfNys3NyMzJyP_-ECghNtlBoQ2Ztagx9TI0W3U5drA5dXU1ANVAWBNQAFoLCR-mVLb9M1KtEANg_bz5wQQyDQoLCgAo9LGSiJPd5AI\"],null,[]],[\"com.instagram.android\",7]]",null,"generic"]]]
```

To implement the `start` parameter I can slice the return from the given `start` and desired `num` parameters, but we will have to fetch every page, every time, due to the token on `batchexecute` (correct me here if I'm wrong).

The `start` parameter was also removed from the `.list` method due to the same behavior

icarcal:
@facundoolano how about return the next token with the reviews response?
This way we can mimic the last behavior of this method, allowing the user to fetch page by page by sending the `page`, `num` and `token` parameter.
The user of the package will be allowed to throttle the request by itself as well, other than only with the package.
It will probably be another breaking change, but this version and the next one will be fully functional

Issue #348, opened 2019-09-20
review's date returns 'Invalid Date'


chihara:
* Operating System: Ubuntu 16.04.2 LTS
* Node version: v12.10.0
* google-play-scraper version: 7.0.1

### Description: 
One review's date returns 'Invalid Date' . not parsed.

I checked google play store's data. This review has Date data was 1 length array.
```
    [
      "gp:AOqpTOH8ZUv5em6TQwWMbnKqPtEWngJ9Ts-S_PsVycbHvtRhL0qIPRzrO8_TVHodY5-v-0W1GRb3kiObyc9oHw",
      ["K. Bacon",[null,2,null,[null,null,"https://lh3.googleusercontent.com/-bsw62Zcu8JI/AAAAAAAAAAI/AAAAAAAAAAA/ACHi3rfRA9WjZvNztK8jPCa2rWsuf99Uyw/photo.jpg"]]],
      4,
      null,
      "概ね良好です。経歴などxxxxxと同期していただければ、有難いです。",
      [1554881170],
      7,
      null,
      null,
      ["101024111675687700835","K. Bacon",null,[[null,2,null,[null,null,"https://lh3.googleusercontent.com/-bsw62Zcu8JI/AAAAAAAAAAI/AAAAAAAAAAA/ACHi3rfRA9WjZvNztK8jPCa2rWsuf99Uyw/photo.jpg"]],true],[null,2,null,[null,null,"https://lh3.googleusercontent.com/Yq7oyNIvAAkuc69fG51sbAQS4otJxbObbt3xdr8tXxXyUdq4tVGtfgeKuptveGdP1srxaHVrNPzOYcfaEQ"]]],
      "2.8.1",
      null,
      [[["vaf_never_display_make_you_happy"],["vaf_never_display_content_language"],["vaf_never_display_app_description"]]]
    ]
```

### Example code:
```js
const playStore = require('google-play-scraper');
var requestInfo = {appId: 'jp.co.recruit.rikunabihaken.rikunabihaken', lang: 'ja', country: 'jp'};
playStore.reviews(requestInfo).then(console.log);
```

### Error message:
date is 'Invalid Date'.
```
  {
    id: 'gp:AOqpTOH8ZUv5em6TQwWMbnKqPtEWngJ9Ts-S_PsVycbHvtRhL0qIPRzrO8_TVHodY5-v-0W1GRb3kiObyc9oHw',
    userName: 'K. Bacon',
[...]

PR #349, opened 2019-09-22
fix 'Invalid Date' on reviews response

icarcal:
#348 

`.reviews`

The date array for reviews dates sometimes have only one position, so I have added `000` as default when the second position is not found.


marxsk:
@icarcal Wouldn't it better to just return undefined? This can introduce new types of errors as people expect that the date is correct. So, they won't detect this immediately.

facundoolano:
Will this return a valid date in all cases?

facundoolano:
Also, does the example in README reflect the format the dates are being returned in?

icarcal:
@facundoolano 
> Will this return a valid date in all cases?

I checked for the example on #348 example and I have the following cases for the date array:
```
[ 1563785933, 50000000 ]
[ 1561903055, 117000000 ]
[ 1558104094, 799000000 ]
[ 1557645130, 99000000 ]
[ 1555736420, 61000000 ]
[ 1554881170 ]
```

So I put together the first part plus the first 3 digits of the second one, which gives me a valid date (it is the same as the store).
The last date do not have the second part, so I add the `000`, for the same reason.
In fact, if you add `000` for all dates, it gives you pretty much the same result, for example:
```
new Date(1563785933000)
// Mon Jul 22 2019 05:58:53 GMT-0300 (Brasilia Standard Time)
new Date(1563785933500)
// Mon Jul 22 2019 05:58:53 GMT-0300 (Brasilia Standard Time)
```

I made some tests and this last part of the timestamp is the milliseconds.
The date will not be changed at all.

I checked for other results, but couldn't find, so I think this covers all cases.

> Also, does the example in README reflect the format the dates are being returned in?

I have updated the README now.
I can change the return to be the given date (`2019-07-22T08:58:53.000Z`) instead the string, what do you think about it?

facundoolano:
@icarcal yes, a Date object would be better if possible

Issue #345, opened 2019-09-14
Some Collections Doesn't Work


vitaliy901:
Some Collections Doesn't Work: 

- topselling_new_free
- topselling_new_paid
- movers_shakers

`gp.list({category: 'GAME', collection: 'topselling_new_paid'}).then(res => {
    console.log(res);
})`

`(node:13656) UnhandledPromiseRejectionWarning: Error: The collection is invalid for the given category or top apps` 

I checked it for every category, but nothing still works ...

icarcal:
Hey @vitaliy901 
I checked for the top apps and these collections does not exist.
You can see it here:
https://play.google.com/store/apps/top/category/GAME?hl=en&gl=us

But, right now just for the GAME and FAMILY related categories, we have the following url (for the new collections):
https://play.google.com/store/apps/new/category/GAME?hl=en&gl=us

BUT, this does not work with other categories, for example:
https://play.google.com/store/apps/new/category/BUSINESS?hl=en&gl=us

Going to take a better look tonight and try to implement this into the `.list` method

vitaliy901:
@icarcal thanks

icarcal:
Hey @vitaliy901 the PR #350 is now openfor review and add support for the `new` collections.
Thanks.

sayginburak:
also facing the same problem. what is the status of the MR? @icarcal 
if it will not be merged soon I will fork, merge and use the package from github. thanks.

icarcal:
@sayginburak the #350  is still under review, you can check the whole discussion there.

fotisskal:
@icarcal do we have any update on this?

I also seem to get the same error "The collection is invalid for the given category or top apps" for the combination
        "category": "events",
        "collection": "trending",
        "country": "id",
        "number": "20"
Is it related?

icarcal:
@fotisskal #350 still under review, I have no update either.

I tested the version 7.0.4 with the data that you sent, and I was able to get the right response.
No error was raised.

What version are you using?

```js
const collection = gplay.collection.TRENDING;
gplay.list({
  collection,
  category: gplay.category.EVENTS,
  country: 'us',
  num: 100
}).then(console.log);
```

Issue #209, opened 2018-04-08
Random test failures


facundoolano:
Specially the `should fetch multiple pages of distinct results` ones. Fix to avoid errors or remove those tests altogether.

icarcal:
Hey @facundoolano can we close this one since the tests have already been fixed?

facundoolano:
yeap

PR #350, opened 2019-09-22
add suport for 'top new' collections

icarcal:
This PR fixes the following issue:
#345

`.list`

This change add support for the `top new` collections, since the url for top apps and new apps are different right now.
I've added support to the following URLs:
- https://play.google.com/store/apps/new
- https://play.google.com/store/apps/new/category/{CATEGORY}

I've also added some validations, since the URL for categories only works for FAMILY and GAME related ones.


icarcal:
The failing test has nothing to do with this PR and is ramdonly failing (#209), probably if you run the tests again it will pass.

lib/list.js
```diff
@@ -41,23 +47,27 @@ function validate (opts) {
   if (opts.age && !R.contains(opts.age, R.values(c.age))) {
     throw Error(`Invalid age range ${opts.age}`);
   }
+}
+
+function defineCollectionUrl (opts) {
```
facundoolano: instead of adding this sort of logic, can you update the object in constants to already contain the path that needs to be called?

lib/utils/helpers.js
```diff
@@ -0,0 +1,16 @@
+const c = require('../constants');
```
facundoolano: I'm not a fan of this sort of generic "helper" modules (specially when this has a single function). This function seems to belong together with the objects in the constants module (which could be renamed to reflect that common responsibility.

icarcal:
@facundoolano I removed the helper and put everything into the constants file as you suggested.

I also tried to keep as close as possible to the collection object that the module already had, so it would not be another breaking change.

lib/utils/parseCategoryApps.js
```diff
@@ -18,8 +18,23 @@ function checkIfCategoryIsReallyPaid (selectedCollection, allCategories) {
   return !firstApps[0].free;
 }
 
-function getParsedCluster (categoryObject, collection) {
-  const collections = {
+function getClusterMappings (collection) {
+  if (c.paths[collection].includes(c.clusters.new)) {
+    return {
+      2: {
```
facundoolano: these objects look like they could be constants in the constants module or in a mappings module, rather than being generated dynamically inside this function.

lib/constants.js
```diff
@@ -1,5 +1,36 @@
...
 
+const { CLUSTER_BASE_URL } = require('./utils/configurations');
+
+const clusters = {
+  new: 'new',
+  top: 'top'
+};
+
+const collectionPaths = {
+  TOP_FREE: { collection: 'topselling_free', path: `${CLUSTER_BASE_URL}/${clusters.top}` },
```
facundoolano: I find the code in this module very hard to follow: 

1. It's hard to tell what the final urls look like by looking at the code here and the code where these constants are used
2. the fact that you need to reduce to export these and then process on the other side to extrac collection + path looks suspicious.
3. this collection+path fields here also seem more than necessary. If you really need to map each category constant to the cluster type (top vs new), then why not

```js
 TOP_FREE: { collection: 'topselling_free', cluster: 'top' },
...
NEW_FREE: { collection: 'topselling_new_free', cluster: 'new' },
... 
```
?

lib/constants.js
```diff
@@ -1,5 +1,36 @@
...
+  TOP_GROSSING_GAMES: { collection: 'topselling_grossing_games', path: `${CLUSTER_BASE_URL}/${clusters.top}` },
+  NEW_FREE: { collection: 'topselling_new_free', path: `${CLUSTER_BASE_URL}/${clusters.new}` },
+  NEW_PAID: { collection: 'topselling_new_paid', path: `${CLUSTER_BASE_URL}/${clusters.new}` },
+  NEW_FREE_GAMES: { collection: 'topselling_new_free_games', path: `${CLUSTER_BASE_URL}/${clusters.new}` },
+  NEW_PAID_GAMES: { collection: 'topselling_new_paid_games', path: `${CLUSTER_BASE_URL}/${clusters.new}` }
+};
+
+module.exports.clusters = clusters;
+
+module.exports.paths = Object.keys(collectionPaths)
```
facundoolano: These look suspicious. We should aim to have very obvious and declarative constants, and then if necessary helper functions to process them as needed when they need to be used.

lib/list.js
```diff
@@ -41,30 +45,30 @@ function validate (opts) {
...
+  const { collection, category } = opts;
+  const path = c.paths[collection];
 
-  opts.lang = opts.lang || 'en';
-  opts.country = opts.country || 'us';
-  opts.num = opts.num || 500;
+  return (category)
```
facundoolano: It's very hard to follow what url is being generated here, and how does it relate to the objects in the constants module.

lib/utils/parseCategoryApps.js
```diff
@@ -18,8 +18,23 @@ function checkIfCategoryIsReallyPaid (selectedCollection, allCategories) {
   return !firstApps[0].free;
 }
 
-function getParsedCluster (categoryObject, collection) {
-  const collections = {
+function getClusterMappings (collection) {
```
facundoolano: I guess I didn't mention when you first introduced this module, but it is solving a rather complex problem and I'm pretty sure it's something that will likely need to update frequently when google rearranges things in the page so we could use a good comment explaining what it does, what clusters are and how this may need to be edited to map what happens in the page.

lib/utils/parseCategoryApps.js
icarcal: I created a new mapping modules for this.

lib/constants.js
icarcal: Definitely agree with you on this.
So I rewrote the constant module to make it cleaner.

lib/utils/parseCategoryApps.js
icarcal: I add some comments and took time to rename some variables to make it more readable.

icarcal:
@facundoolano made some changes to this PR

icarcal:
@facundoolano do you need any other change to this PR?

facundoolano:
Sorry for the delay, I haven't had time to get back to this. I'll try to do it in the next few days.

PR #369, opened 2019-11-18
add types file

Akimyou:
Hello. 

Thank you for  this awesome scraper project first, it help me a lot in my job.
Then i write a ts types file for this project: `index.d.ts`.
Hope you can review and accept it, it make this project more friendly to typescript user.

Thanks.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,327 @@
+// constants
+
+export enum category {
+  APPLICATION = 'APPLICATION',
+  ANDROID_WEAR = 'ANDROID_WEAR',
+  ART_AND_DESIGN = 'ART_AND_DESIGN',
+  AUTO_AND_VEHICLES = 'AUTO_AND_VEHICLES',
+  BEAUTY = 'BEAUTY',
+  BOOKS_AND_REFERENCE = 'BOOKS_AND_REFERENCE',
+  BUSINESS = 'BUSINESS',
+  COMICS = 'COMICS',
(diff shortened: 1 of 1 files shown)


facundoolano:
Full disclosure: I never used TypeScript. 

I'm ok with adding this, but I'm worried about the fact that the scraper is constantly changing and this file could easily get out of sync.

Is there any way we can lint/test these definitions so we get a travis error when they go out of sync?

Akimyou:
Sorry, don't have an way lint/test the definitions. We have to do it manually when it out of sync.

There are two way to public a types file.
- public as an @types package to: http://definitelytyped.org/
- public as an file to original repo

Both two ways have to sync it manually. 😭

dasheck0:
What is the current status of this? Need help with that? I'd also like some type definitions. So I would be glad to help out.

facundoolano:
@dasheck0 status is pretty much the same as in my first comment. I'm ok with merging this but I won't maintain type definitions, nor can I make other contributors maintain them unless there's some sort of test/linter that does it as part of the PR build.

Hazealign:
@Akimyou Thanks for awesome works. This PR will make library more kindly for TypeScript users.

@facundoolano Thanks for awesome library too. It saves my lot of works. Unfortunately we can't confirm the type definition automatically like lint or test tools. But I think if this PR got merged, this project's type definition get contributed well. Because this project has nice documents. So other many TypeScript users(includes me) can improve type definitions from documents.

PR #372, opened 2019-11-22
Avoid exception when the app doesn't have reviews

mmoksh:
Check this app:

https://play.google.com/store/apps/details?id=air.vjsol.macheV.live

The app doesn't have any reviews in English yet.

Instead of returning an empty list for the following API call, it throws an exception: "TypeError: Cannot read property 'map' of undefined".

`gplay.reviews({
	appId: "air.vjsol.macheV.live",
	num: 2
}).then(console.log, console.log);`

This PR fixes the issue.


mmoksh:
@facundoolano thanks for merging this PR. Would you please update the package in the npm repository?

Issue #378, opened 2019-12-17
Add new field if app is avaliable in this geo


yurganov:
* Operating System: macOS
* Node version: 12.4.0
* google-play-scraper version: 7.1.1

### Description: 
There are some apps that does not avaliable in some geo.

For example, 

https://play.google.com/store/apps/details?id=com.iqoption&gl=us is not available in US,
but is available in Russia: https://play.google.com/store/apps/details?id=com.iqoption&gl=ru

If app is available, there is «Install» button.

Let's add some field that can figure out, if app is available or not.

yurganov:
Fixed in https://github.com/facundoolano/google-play-scraper/pull/379, need to approve way of fixing

PR #379, opened 2019-12-17
add new field isAvailable

yurganov:
This is fixing https://github.com/facundoolano/google-play-scraper/issues/378

@facundoolano what you think about modifying parse function in such way?

I did not found any other place where I can get raw HTML to parse it.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -136,7 +136,8 @@ const MAPPINGS = {
   editorsChoice: {
     path: ['ds:5', 0, 12, 15, 0],
     fun: Boolean
-  }
+  },
+  available: ['internal:available']
 };
 
 function descriptionText (description) {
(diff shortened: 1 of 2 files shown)


yurganov:
@facundoolano take a look please

lib/app.js
```diff
@@ -136,7 +136,8 @@ const MAPPINGS = {
   editorsChoice: {
     path: ['ds:5', 0, 12, 15, 0],
     fun: Boolean
-  }
+  },
+  available: ['internal:available']
```
facundoolano: I'm not convinced by this syntax. If a field doesn't fit with the MAPPINGS structure add it separately in the parsing code (with an explanatory command) but not here, since it's not really meant to be extracted from the same data object

facundoolano:
Closing due to inactivity.

Issue #380, opened 2019-12-26
Add currency and price to method list


yurganov:
* Operating System: macOS
* Node version: 12.4.0
* google-play-scraper version: 7.1.1

### Description: 
I am using list method to collect information about free/paid/grossing top.

There is no fields «price» and «currency». Let's fix this.

### Example code:
```js
app.get('/top/android', function (req, res) {
    let data = {
        category: req.query.category_name || null,
        collection: req.query.collection || "topselling_free",
        num: req.query.num || 1000,
        country: req.query.gl || "us",
        lang: req.query.hl || "en",
        directCollection: req.query.direct || false,
    };

    gplay.list(data)
        .then(
            result => res.send(JSON.stringify(result)),
            result => res.send(JSON.stringify(result))
        );
});
```

### Error message:

None

facundoolano:
fixed

PR #381, opened 2019-12-26
add price and currency field to mappings of appList

yurganov:
Fix of https://github.com/facundoolano/google-play-scraper/issues/380

### DIFF
--- a/lib/utils/appList.js
+++ b/lib/utils/appList.js
@@ -21,6 +21,14 @@ const MAPPINGS = {
     path: [7, 0, 3, 2, 1, 0, 2],
     fun: (price) => price === undefined ? 'FREE' : price
   },
+  currency: {
+    path: [7, 0, 3, 2, 1, 0, 2],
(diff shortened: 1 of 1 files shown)


lib/utils/appList.js
```diff
@@ -21,6 +21,14 @@ const MAPPINGS = {
     path: [7, 0, 3, 2, 1, 0, 2],
     fun: (price) => price === undefined ? 'FREE' : price
   },
+  currency: {
+    path: [7, 0, 3, 2, 1, 0, 2],
+    fun: (price) => price === undefined ? false : price.match(/([^0-9\.,\s]+)/)[0]
```
facundoolano: why default to false instead of undefined?

facundoolano:
You have an eslint error (See travis build)

lib/utils/appList.js
yurganov: fixed

Issue #311, opened 2019-05-18
fullDetail not working in search


facundoolano:
* google-play-scraper version: 6.2.6

### Description: 
After the recent restructuring of the code, the `fullDetail` option of the search method is not being honored.

We need to implement something similar as done in parseList, where a new app method call is done per each yielded result.

Also add a test proving that full detail works for search.

### Example code:
```js
g.search({term: 'panda', num:1, fullDetail: true }).then(console.log)

[ { title: 'Panda Pop! Bubble Shooter Saga & Puzzle Adventure',
    appId: 'com.sgn.pandapop.gp',
    url: 'https://play.google.com/store/apps/details?id=com.sgn.pandapop.gp',
    icon: 'https://lh3.googleusercontent.com/n-N5SeSksOVootPWoYDGek1iY-_EeoL46hhWOgnxbbP0eypVMqMlJtOwAZSzN-yDoOtO',
    developer: 'Jam City, Inc.',
    developerId: '5509190841173705883',
    priceText: 'FREE',
    free: true,
    summary: 'Aim, match, shoot and pop in this free bubble shooter classic saga adventure!',
    scoreText: '4.4',
    score: 4.4345155 } ]
```

v1shky:
Any progress on this issue by any forks?

efeconirulez:
+1

girotomas:
```
gplay
  .search({
    throttle: 5,
    term: "panda",
    num: 2,
    fullDetail: true
  })
  .then(
    l =>
      l.map(a => gplay.app({ appId: a.appId }).then(console.log, console.log)),
    console.log
  );
```

efeconirulez:
Still no progress?

sgallese:
fullDetail seems to be working for me.

KavarH:
For me it is still not working.
It is just showing a short summary.
Doing what @girotomas wrote you see the difference.

eltriplet:
> This might be useful for someone
> 
> ```
> gplay
>   .search({
>     throttle: 5,
>     term: "panda",
>     num: 2,
>     fullDetail: true
>   })
>   .then(
>     l =>
>       l.map(a => gplay.app({ appId: a.appId }).then(console.log, console.log)),
>     console.log
>   );
> ```

If you want to store the result in a json array, instead of printing on terminal, the following adaptation of @girotomas code worked for me:

var gplay = require('google-play-scraper');

justDoIt();

async function justDoIt() {
[...]

PR #384, opened 2020-01-20
Added fullDetail for search, fixed testcases

KavarH:
This is fixing #311 

@facundoolano is this kind of like you wanted it to be implemented?

In addition, I fixed the failing test cases. I hope in a good way.

I hope I did not do something one does not do anymore for some years, since I'm new to js and needed to google a lot to understand everything (I had some problems with pending promises until I understood where they come from).

### DIFF
--- a/lib/search.js
+++ b/lib/search.js
@@ -1,6 +1,7 @@
 'use strict';
 
 const request = require('./utils/request');
+const app = require('./app');
 const scriptData = require('./utils/scriptData');
 const appList = require('./utils/appList');
 const R = require('ramda');
@@ -22,6 +23,26 @@ function processAndRecur (html, opts, savedApps, mappings) {
   return checkFinished(opts, [...savedApps, ...apps], token);
 }
 
+function getFullDetail (apps, num) {
+  return new Promise(function (resolve) {
+    resolve(app({ appId: apps[num].appId }));
+  });
+}
+
+function details (savedApps, num) {
+  let i;
+  const y = [];
+  let x;
+  for (i = 0; i < num; i++) {
+    x = getFullDetail(savedApps.slice(0, num), i).then(function (result) {
+      y.push(result);
+    }).then(function () {
+      return y;
+    });
+  }
+  return x;
+}
+
 /*
  * If already have requested results or there are no more pages, return current
  * app list, otherwise request the ajax endpoint of the next page and process
@@ -29,7 +50,12 @@ function processAndRecur (html, opts, savedApps, mappings) {
  */
 function checkFinished (opts, savedApps, nextToken) {
   if (savedApps.length >= opts.num || !nextToken) {
-    return savedApps.slice(0, opts.num);
+    if (opts.fullDetail) {
+      let x = details(savedApps, opts.num);
+      return x;
+    } else {
+      return savedApps.slice(0, opts.num);
+    }
   }
 
   const requestOptions = Object.assign({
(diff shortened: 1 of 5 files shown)


lib/search.js
```diff
@@ -28,7 +29,24 @@ function processAndRecur (html, opts, savedApps, mappings) {
  * the results.
  */
 function checkFinished (opts, savedApps, nextToken) {
+  // loads fullDetail for all apps that are being returned (starting with last item)
+  function load (num) {
```
facundoolano: this looks a bit more complicated than it could be in my opinion. Considering we need to call the app method once per each result, I think it would be best to adding something like this at the bottom of the module:

```js
..then((results) => {
    if (opts.fullDetail) {
      return results.map(() => app({...}))
    }
    return results;
  })
```

Also, it's important that when you call the app method you pass the same opts used for the search, so options like language are respected.

test/lib.list.js
```diff
@@ -176,7 +176,7 @@ describe('List method', () => {
         app.screenshots.map(assertValidUrl);
         app.comments.map(assert.isString);
       }));
-  }).timeout(timeout);
+  }).timeout(timeout * 2);
```
facundoolano: this is suspicious, it's probably better to increase the global timeout in the package.json than in the code.

lib/search.js
KavarH: this really looks easier, I'm gonna try this

facundoolano: oh and you probably need to call Promise.all on the resulting list to turn it into alist of results instead of a list of promises

test/lib.list.js
KavarH: suspicious in which  way? The timeout is already defined in the top of the method and only those tests using the fullDetail are having problems with the timeout.
But maybe the change in how to get the full detailed apps resolves this anyway.

lib/search.js
```diff
@@ -122,6 +123,12 @@ function search (getParseList, opts) {
     initialRequest(opts)
       .then(resolve)
       .catch(reject);
+  }).then((results) => {
+    if (opts.fullDetail) {
+      // if full detail is wanted get it from the app module
+      return Promise.all(results.map((app) => appData({ opts, appId: app.appId })));
```
facundoolano:
```suggestion
      return Promise.all(results.map((app) => appData({ ...opts, appId: app.appId })));
```

facundoolano: your version is not setting the right properties

KavarH:
I had some time to test if everything (especially the options) are working now => They are.

However, it's for me not rlly understandable why this CI fails from time to time when trying to get this 55 distinct apps. Would have tried to fix it too, but it does not happen for me when testing it locally.

facundoolano:
I think that error was preexistent; I haven't looked into it so I'm not sure what's the cause either.

KavarH:
Yes it was, just thought about it, since I tried to fix those errors with test cases.

lib/search.js
```diff
@@ -1,6 +1,7 @@
 'use strict';
 
 const request = require('./utils/request');
+const appData = require('./app');
```
facundoolano: one last thing, by requiring the appMethod here, you may be loosing the [memoization option](https://github.com/facundoolano/google-play-scraper#memoization) when it's set.

Instead of requiring you should pass the app function as an argument in index.js, like getParseList is passed now

KavarH: Oh, thats actually something I haven't even thought on.
Going change this.
Thanks

PR #385, opened 2020-01-27
Change example app

gqqnbig:
because the original app is no longer available on Google Play

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,57 +37,64 @@ Example:
 ```javascript
 var gplay = require('google-play-scraper');
 
-gplay.app({appId: 'com.dxco.pandavszombies'})
+gplay.app({appId: 'com.google.android.apps.translate'})
   .then(console.log, console.log);
 ```
 Results:
 
 ```javascript
-{ appId: 'com.dxco.pandavszombies',
-  url: 'https://play.google.com/store/apps/details?id=com.dxco.pandavszombies&hl=en&gl=us'
-  title: 'Panda vs Zombies',
-  description: 'Panda, my friend, Panda is the answer. But not any Panda: Rocky the Panda!',
-  descriptionHTML: 'Panda, my friend, Panda is the answer. But not any Panda: <b>Rocky the Panda!</b>',
-  summary: 'Help Rocky the Panda warrior to fight zombie games and save the Panda kind.',
-  installs: '10,000+',
-  minInstalls: 10000,
-  score: 4.5025907,
+{
+  title: 'Google Translate',
+  description: 'Translate between 103 languages by typing\r\n...' ,
+  descriptionHTML: 'Translate between 103 languages by typing<br>...',
+  summary: 'The world is closer than ever with over 100 languages',
+  installs: '500,000,000+',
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,57 +37,94 @@ Example:
...
-  title: 'Panda vs Zombies',
-  description: 'Panda, my friend, Panda is the answer. But not any Panda: Rocky the Panda!',
-  descriptionHTML: 'Panda, my friend, Panda is the answer. But not any Panda: <b>Rocky the Panda!</b>',
-  summary: 'Help Rocky the Panda warrior to fight zombie games and save the Panda kind.',
-  installs: '10,000+',
-  minInstalls: 10000,
-  score: 4.5025907,
+{
+  title: 'Google Translate',
+  description: '• Translate between 103 languages by typing\r\n' +
```
facundoolano: why those '+' between strings? this is supposed to be the actual output returned by the node shell, and if it's too long just cut it or add a ...

README.md
```diff
@@ -37,57 +37,94 @@ Example:
...
-  version: '1.4',
-  recentChanges: '- Added a hint system<br>- Added share option in level finished',
-  comments: [ 'Great!', 'PvZ', 'LOL', 'Zombie killer', 'Una pasada' ]
-  }
+  icon: 'https://lh3.googleusercontent.com/ZrNeuKthBirZN7rrXPN1JmUbaG8ICy3kZSHt-WgSnREsJzo2txzCzjIoChlevMIQEA',
+  headerImage: 'https://lh3.googleusercontent.com/e4Sfy0cOmqpike76V6N6n-tDVbtbmt6MxbnbkKBZ_7hPHZRfsCeZhMBZK8eFDoDa1Vf-',  screenshots: [
+    'https://lh3.googleusercontent.com/dar060xShkqnJjWC2j_EazWBpLo28X4IUWCYXZgS2iXes7W99LkpnrvIak6vz88xFQ',
+    'https://lh3.googleusercontent.com/VnzidUTSWK_yhpNK0uqTSfpVgow5CsZOnBdN3hIpTxODdlZg1VH1K4fEiCrdUQEZCV0',
+    'https://lh3.googleusercontent.com/zo76JEdXGsBcxbsJXX3XPN0pGicjIYAy7mi95al8LTT14H4xrer-lE2kR_DEXCn-GY_l',
+    'https://lh3.googleusercontent.com/8HwziIPNKGnrJe5l6MfdAm6XeZ7nYEWjLXli34D5xyzBT9jE1qIQstzZXJP9X63b4A',
```
facundoolano: same with this, this object is expected to be illustratory, there's no benefit in showing 20 images vs 4, just makes the readme longer

README.md
gqqnbig: cmd_2020-01-27_12-52-13

To answer your first question, it is the app method that gives the + strings, unless I'm using the API in a wrong way.

But I agree the example will be lengthy. Let me change it.

gqqnbig:
I made a forced push. See if the new changes look better for you. Thanks.

Issue #387, opened 2020-02-08
Cannot read property '4' of null in gplay.app


gabidabet:
* Operating System:Windows 10
* Node version: v10.15.0
* google-play-scraper version: last version

### Description: 
I get that error for somme app ids like :
-air.com.peppapig.paintbox "the exemple"
-com.netmarble.mherosgb
-com.wb.goog.injustice

### Example code:
```js
// Put code to reproduce the issue here
gplay
    .app({
        appId: "air.com.peppapig.paintbox"
    })
    .then(console.log, console.log);

### Error message:
Cannot read property '4' of null

**Comment by s-h-a-d-o-w** [comment]
Same for `.list`, e.g.:
```javascript
gplay
  .list({
    collection: gplay.collection.TOP_PAID_GAMES,
    fullDetail: true,
    num: 2,
    throttle: 10, // Max. 10 requests per second, avoid Google's limit
  })
  .then(console.log, console.log);
```

I suppose Google changed the structure of the site? :/

icarcal:
@gabidabet and @s-h-a-d-o-w thanks for reporting.
Yes, gplay changed the structure.

I'll send a PR to fix this, ASAP

facundoolano:
This should be fixed now.

cgwyllie:
@facundoolano / @icarcal / @ We've been seeing this error coming up in our logs recently.

https://github.com/facundoolano/google-play-scraper/pull/409 was merged 2 days ago which seems to have changed the mappings for comments from `ds:18` to `ds:17`.

The `ds:18` field seemed to be part of the original fix for this issue in: https://github.com/facundoolano/google-play-scraper/pull/389

Has this regressed or is the store returning multiple different shapes for different app listings?

Any help here would be appreciated.

icarcal:
@cgwyllie the mapping is correct for countries other than `us`

Specifically for `us`, google play is returning a string 'er' at the position `ds:17` and the correct comments at `ds:19`

Check the examples for the same app in different countries:
<details>
<summary>.app with multiple countries example</summary>

```javascript

gplay.app({
  appId: 'com.facebook.katana'
}).then(console.log)

gplay.app({
  appId: 'com.facebook.katana',
  country: 'us',
  lang: 'en'
}).then(console.log)

gplay.app({
  appId: 'com.facebook.katana',
  country: 'br',
  lang: 'pt-BR'
}).then(console.log)

gplay.app({
  appId: 'com.facebook.katana',
[...]

PR #389, opened 2020-02-10
fix error "Cannot read property '4' of null in gplay.app"

icarcal:
This PR fixes the following issue:
#387 

`.app`

This change fix the error "Cannot read property '4' of null in gplay.app".
The structure of the arrays at gplay store have changed, so I just updated the MAPPINGS variable

I have added some debugging too.

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -186,7 +186,7 @@ function extractComments (comments) {
     return [];
   }
 
-  debug('comments: %O', comments)
+  debug('comments: %O', comments);
 
   return R.compose(
     R.take(5),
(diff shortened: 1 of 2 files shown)


icarcal:
The tests are failing due to other issues, unrelated to this one.
I'm going to send another PR to fix the tests.

facundoolano:
There's a style error which is from the new code. Also, can you update the app test so it breaks for the no comments case?

icarcal:
@facundoolano I've rebased my branch with the latest updates in the `dev` branch, and fixed the style error.
Now the errors are related to the store itself.
I'm going to take a better look and see if I can fix those tests in another PR.

PR #412, opened 2020-06-21
fix .app comments

icarcal:
This PR fix the following issues:
[comment](https://github.com/facundoolano/google-play-scraper/issues/387#issuecomment-646092309)

`.app`

Mappings for the comments are randomly being put in different places:
 - e.g: `country br -> ds:17`
 - e.g: `country es -> ds:18`
 - e.g: `country us -> ds:19`

To solve this I have mapped AF_dataServiceRequests from the html, and there we have an id
 - e.g: `{ 'ds:17' : {id:'UsvDTd',ext: 1.36880256E8 ,request:[null,null,[2,null,[40]]]}`

This id does not change between countries and is unique for each mapping, so it is a safe way to go, other than map this for each country.

I've also done the following:
- [0b68285](https://github.com/facundoolano/google-play-scraper/commit/0b68285781c612eaae812624cc316b9cc4231050) Added husky as dev dependency to prevent local errors to be pushed to the remote;
- [24e46ab](https://github.com/facundoolano/google-play-scraper/commit/24e46abcc19c31f8cdf32e139538daa969a2a647) Allow the `rawInput`(parsedData) to be passed as an argument for `.fun` function on mappings
- [19da778](https://github.com/facundoolano/google-play-scraper/commit/19da778c4b4882bdf788863e98f631f1c7404d6b) Added a test for non default language and validation for comment array
- [ca4fd70](https://github.com/facundoolano/google-play-scraper/commit/ca4fd707f88e1228698e284d7fa0bf1bda841161) Remove distinct validation tests for `.search` method, since google play seems to duplicate some of the apps in the response randomly (you can check this at the gplay store)

<details>
<summary>Google Play Store duplication example</summary>
Screen Shot 2020-06-20 at 21 24 41

Screen Shot 2020-06-20 at 21 51 29

</details>

### DIFF
--- a/lib/mappers/details.js
+++ b/lib/mappers/details.js
@@ -129,17 +129,24 @@ function buildHistogram (container) {
   };
 }
 
-function extractComments (comments) {
-  if (!comments) {
+// Google Play responds with the country us and lang en
+// at ds:17 a single string 'er', but for other languages and countries
+// it responds correctly
+function extractComments (comments, rawInput) {
+  const fallbackComments = (!R.is(Array, comments))
+    ? R.path(['ds:19', 0], rawInput)
+    : comments;
+
+  if (!fallbackComments) {
     return [];
   }
 
-  debug('comments: %O', comments);
+  debug('comments: %O', fallbackComments);
 
   return R.compose(
     R.take(5),
     R.reject(R.isNil),
-    R.pluck(4))(comments);
+    R.pluck(4))(fallbackComments);
 }
 
 module.exports = MAPPINGS;
--- a/lib/mappers/details.js
+++ b/lib/mappers/details.js
@@ -89,7 +89,10 @@ const MAPPINGS = {
   version: ['ds:8', 1],
   recentChanges: ['ds:5', 0, 12, 6, 1],
   comments: {
+    // NON US APPS COMMENT SECTION
     path: ['ds:17', 0],
+    // US APPS COMMENT SECTION
+    fallback: ['ds:19', 0],
     fun: extractComments
   },
   editorsChoice: {
@@ -129,24 +132,21 @@ function buildHistogram (container) {
   };
 }
 
-// Google Play responds with the country us and lang en
-// at ds:17 a single string 'er', but for other languages and countries
-// it responds correctly
-function extractComments (comments, rawInput) {
-  const fallbackComments = (!R.is(Array, comments))
-    ? R.path(['ds:19', 0], rawInput)
-    : comments;
-
-  if (!fallbackComments) {
+/**
+ * Extract the comments from google play script array
+ * @param {array} comments The comments array
+ */
+function extractComments (comments) {
+  if (!comments) {
     return [];
   }
 
-  debug('comments: %O', fallbackComments);
+  debug('comments: %O', comments);
 
   return R.compose(
     R.take(5),
     R.reject(R.isNil),
-    R.pluck(4))(fallbackComments);
+    R.pluck(4))(comments);
 }
 
 module.exports = MAPPINGS;
--- a/lib/mappers/details.js
+++ b/lib/mappers/details.js
@@ -89,10 +89,8 @@ const MAPPINGS = {
   version: ['ds:8', 1],
   recentChanges: ['ds:5', 0, 12, 6, 1],
   comments: {
-    // NON US APPS COMMENT SECTION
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -26,6 +26,7 @@
     "cheerio": "^0.22.0",
     "debug": "^2.2.0",
     "eslint": "^5.16.0",
+    "husky": "^4.2.5",
```
facundoolano, line 29: this should be a dev dependency

package.json
```diff
@@ -42,5 +43,11 @@
     "mocha": "^5.2.0",
     "promise-log": "^0.1.0",
     "validator": "^5.2.0"
+  },
+  "husky": {
+    "hooks": {
+      "pre-commit": "npm run lint",
+      "pre-push": "npm run test"
```
facundoolano, line 50: this may be a bit too much to demand, unfortunately the scraper will always break from time to time, and this would prevent pushing legitimate fixes when unrelated tests fail

lib/mappers/details.js
```diff
@@ -129,17 +129,24 @@ function buildHistogram (container) {
   };
 }
 
-function extractComments (comments) {
-  if (!comments) {
+// Google Play responds with the country us and lang en
```
facundoolano, line 132: Have you verified this by code or just in the browser? I've noticed that when logged in the browser the mappings differ, could that be related?

I find it suspicious that this only happens for this field, I wonder if there are other scenarios we're missing

lib/mappers/details.js
```diff
@@ -129,17 +129,24 @@ function buildHistogram (container) {
...
+// at ds:17 a single string 'er', but for other languages and countries
+// it responds correctly
+function extractComments (comments, rawInput) {
+  const fallbackComments = (!R.is(Array, comments))
+    ? R.path(['ds:19', 0], rawInput)
```
facundoolano, line 137: I wonder if we could incorporate this to the mappings somehow, I guess it's ok if this is a one-time thing for this specific field, but this could easily go out of sync next time mappings change

lib/mappers/details.js
icarcal: I looked into google play on an anonymous tab, and `ds:17` is the only field that has the `isError` attributte set to true, and **only** when I add `hl=en&gl=us` as url parameters.
```javascript
<script nonce="">AF_initDataCallback({key: 'ds:17', isError:  true ,  data:["er"]});</script>
```

When I logged in google play, the only field with `isError` true is `ds:24`:
```javascript
<script nonce="">AF_initDataCallback({key: 'ds:24', isError:  true ,  data:["er"]});</script>
```

The mappings for comments also change, `ds:27` for `us/en` and `ds:25` for any other country/lang

package.json
icarcal: You can [bypass](https://github.com/typicode/husky/issues/124#issuecomment-299847973) husky(git hooks) by adding --no-verify to the git command.
i.e: `git commit --no-verify` and git `push --no-verify`

package.json
icarcal: My bad, gonna fix this

lib/mappers/details.js
icarcal: How about [this](https://github.com/facundoolano/google-play-scraper/pull/412/commits/d83a3efdb9e3a2695ee63c8b9d44d3fccd69dced)?

facundoolano: yeah that seems better, just add a comment explaining why that fallback is sometimes necessary (and what that 'er' vaue means)

icarcal: done

icarcal:
@facundoolano I have updated the PR description after taking a better look at #420  and [this feedback](https://github.com/facundoolano/google-play-scraper/issues/387#issuecomment-652631039)

The real cause was the recent changes of google play and the addition of the popular characters section:
Screen Shot 2020-07-23 at 02 13 19

So with this, mappings for the comments are randomly being put in different places:
 - e.g: `country br -> ds:17`
 - e.g: `country es -> ds:18`
 - e.g: `country us -> ds:19`

To solve this I have mapped AF_dataServiceRequests from the html, and there we have an id:
Screen Shot 2020-07-23 at 04 23 16

This id does not change between countries and is unique for each mapping and between countries, so it is a safe way to go, other than map this for each country.

PR #423, opened 2020-08-01
Fix .reviews method

mmoksh:
Let's say we need the most recent 10 reviews for the Instagram app. The .reviews method will try to load all reviews for the app (millions) then return the required 10 reviews.

This PR will stop the HTTP requests once the library has loaded the required reviews.


mmoksh:
@facundoolano Would you please update the code on the npm repository as well? https://www.npmjs.com/package/google-play-scraper

PR #424, opened 2020-08-02
Fix .reviews method

mmoksh:
The .reviews method returns the reviews in the last page only and discard the previously fetched reviews. E.g. if we request 1000 reviews, the library will fetch +1000 reviews, but will return only 150 reviews.

This PR fixes the issue.


mmoksh:
@facundoolano when you merge this, would you please update the npm repository as well? https://www.npmjs.com/package/google-play-scraper

PR #426, opened 2020-08-09
Fix response when app doesn't have reviews

mmoksh:
The .reviews function crashes when trying to get the reviews of this app:

https://play.google.com/store/apps/details?id=com.knifedesignideas.pondokonline7

The "formatReviewsResponse" expects a json object of three keys as input, but it receives input of different type.

This PR fixes the issue.


mmoksh:
@facundoolano I updated the npm version so that it's easier to publish on the npm repository.

PR #462, opened 2021-03-08
add reviews sortType

majidfarzaneh:
add reviews sortType


facundoolano:
I think you mistaken the repository, that looks like google-play-api code.

PR #494, opened 2021-11-24
Changed HTTP library, updated tests

VitaliyRDev:
Changed HTTP library from 'request' to 'got', updated tests, also implemented new throttle function for the compatibility with the 'got' library.

### DIFF
--- a/lib/utils/sleep.js
+++ b/lib/utils/sleep.js
@@ -1,0 +1,3 @@
+module.exports = function (ms) {
+    return new Promise(resolve => setTimeout(resolve, ms));
+  };
--- a/lib/utils/sleep.js
+++ b/lib/utils/sleep.js
@@ -1,3 +1,0 @@
-module.exports = function (ms) {
-    return new Promise(resolve => setTimeout(resolve, ms));
-  };
--- a/lib/utils/throttle.js
+++ b/lib/utils/throttle.js
@@ -2,21 +2,37 @@ function sleep (ms) {
   return new Promise(resolve => setTimeout(resolve, ms));
 };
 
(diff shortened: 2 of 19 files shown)


lib/utils/sleep.js
```diff
@@ -0,0 +1,3 @@
+module.exports = function (ms) {
```
facundoolano, line 1: a module for this is overkill, just duplicate the function in the two modules that use it

lib/utils/throttle.js
```diff
@@ -0,0 +1,52 @@
+const sleep = require('./sleep.js');
+
+function Throttle () {
```
facundoolano, line 3: is this your code or you copied or adapted it from somewhere else?

VitaliyRDev: Throttle function I've written by myself, sleep function is copied

facundoolano: Good, in that case, can you add a docstring explaining how this works (e.g. expected settings) and some basic unit tests showing that it works?

lib/utils/sleep.js
VitaliyRDev: > 

Done.

lib/utils/throttle.js
VitaliyRDev: Sure, but it will take some time.

facundoolano: I have no hurry, thank you for the contribution.

VitaliyRDev:
@facundoolano waiting for review...

facundoolano:
Thanks!

PR #511, opened 2022-01-16
Simplify code to find collection clusters for list method

facundoolano:


facundoolano:
@icarcal I would appreciate your thoughts on this one. It seems that we had defensive code for the case where google changes the amount of clusters returned _and_ the collections change order depending on that amount of clusters. Not sure how frequent that is, I would prefer to default to assuming the same order and see if test breaks or we spot an error.

facundoolano:
On a closer look, I wasn't taking into account that setting a category changes the list of clusters as well. And by choosing categories I see that, for instance, a TRENDING cluster does appear sometimes (I was removing it because I didn't see it in the global collections).

But still the current implementation doesn't seem accurante, I see for example TOP FREE/TOP GROSSING/TRENDING for EVENTS category, which doesn't match our [current patterns for 3 clusters](https://github.com/facundoolano/google-play-scraper/blob/daa093537a046b093f78f49b14a02c99ef24c793/lib/mappers/clusters.js#L12-L15) (which is TOP FREE/TOP PAID/GROSSING)
image

facundoolano:
I changed the implementation to use the collection name (e.g. `top grossing games`) to identify the proper cluster, instead of a list of positions.

lib/list.js
```diff
@@ -49,18 +50,20 @@ function validate (opts) {
...
 
-function buildUrl (opts) {
+function buildInitialUrl (opts) {
   const url = defineCollectionUrl(opts);
 
+  // For the initial collection lookup we force the language to english
+  // since we depend on collection names to identify the proper cluster later on
   const queryString = {
-    hl: opts.lang,
+    hl: 'us',
```
icarcal, line 66: This wouldn't give us wrong collections depending on the language that is being sent?

lib/list.js
```diff
@@ -75,4 +78,84 @@ function buildUrl (opts) {
   return fullURL;
 }
 
+const COLLECTIONS_MAPPING = ['ds:3', 0, 1];
+const CLUSTER_URL_MAPPING = [0, 0, 3, 4, 2];
+const CLUSTER_NAME_MAPPING = [0, 1];
+const CLUSTER_NAMES = {
+  TOP_FREE: 'top free apps',
```
icarcal, line 85: Gotcha, you make the first request in English to get the cluster names, right?

facundoolano: That's right, I checked and the clusters are the dale regardless of language.

lib/list.js
facundoolano: I added a test for that case, the cluster is selected properly in English and then the list is parsed in whatever language you requested

PR #512, opened 2022-01-17
reorganize modules

facundoolano:
This removes concepts of mappers and requesters from modules, basically making each public method self-contained in its own module. The only utils kept around are those where we are actually reusing code in more than one place (and those cases will be revisited in separate PRs).

There still some duplication of the processAndRecur/processPages functions of the different methods, which should also be refactored in other PRs if possible.


lib/requesters/appsMappedRequests.js
```diff
@@ -32,19 +40,10 @@ async function processFullDetailApps (apps, opts) {
...
-
-  const processedApps = appList.extract(mappings.apps, html);
-  const apps = opts.fullDetail
-    ? await processFullDetailApps(processedApps, opts)
-    : processedApps;
-  const token = R.path(mappings.token, html);
-
-  return checkFinished(opts, [...savedApps, ...apps], token);
-}
+const REQUEST_MAPPINGS = {
```
icarcal, line 43: Do you believe that adding this closer to where this is used will be better for maintenance?

facundoolano: yes I want to reduce module surface (fewer deeper modules instead of multiple shallower ones), especially for things that are used in a single place I'd rather have them in a single spot. 

(I haven't finished this PR btw)

Issue #495, opened 2021-12-02
can't get kr rankings.


berrybar:
* Operating System: windows 10
* Node version: 10.16.0
* google-play-scraper version: 8.0.4

### Description: 
When I tested in the US and Japan, I was able to get the value.
But when I set it to Korea, I couldn't get the value at all.
It hasn't been since December 1st in Korea, and it seems to have found a problem.
It seems that the Korean Google Play ranking collection url is pointing to the wrong path.
Can I get feedback on this?

### Example code:
```js
await G.list({
            category: G.category.GAME,
            collection: collection,
            num: 200,
            lang: 'ko',
            country: 'kr',
            throttle: 10
        })
```

### Error message:

```
Error: The collection topgrossing is invalid for the given category, top apps or new apps
```

ansoncen:
What I found in the region kr, hk and tw, app detail page layout upgrade to the new version.  The lib's parse function should be updated so.

berrybar:
I agree with that opinion.

druphliu:
@VitaliyRDev

13389203735:
> on 7 Dec 2021

I also encountered this problem. How did you solve it? Please tell me

13389203735:
> 我在kr、hk和tw地区发现，app详情页布局升级到新版本。lib的解析
我也出现了这个问题，你是怎么解决的，麻烦告知一下

13389203735:
> What I found in the region kr, hk and tw, app detail page layout upgrade to the new version. The lib's parse function should be updated so.

I use the dev branch code locally, but I found that the app store can't get information in Kr, HK and tw

I also have this problem. How did you solve it? Please tell me

13389203735:
@ansoncen I use the dev branch code locally, but I found that the app store can't get information in Kr, HK and tw

I also have this problem. How did you solve it? Please tell me

berrybar:
> > on 7 Dec 2021
> 
> I also encountered this problem. How did you solve it? Please tell me

I also haven't solved it yet.

icarcal:
Looks like google play completely changed the kr store as @ansoncen pointed out.
The URL used by the lib is redirecting to `/apps`:
`https://play.google.com/store/apps/top/category/GAME/?hl=ko&gl=kr`

I'm not sure if this new store version has the `top` or `new` URL anymore.

Apparently, it does have some way to get the top/new apps, but differently than we do today:
[...]

PR #523, opened 2022-02-25
Added Special Mappings to be used for the new store design (used in H…

Djip:
Fixing #495 in terms of using the `.search` feature


lib/search.js
```diff
@@ -8,6 +8,7 @@ const { BASE_URL } = require('./constants');
 
 const body = 'f.req=%5B%5B%5B%22qnKhOb%22%2C%22%5B%5Bnull%2C%5B%5B10%2C%5B10%2C50%5D%5D%2Ctrue%2Cnull%2C%5B96%2C27%2C4%2C8%2C57%2C30%2C110%2C79%2C11%2C16%2C49%2C1%2C3%2C9%2C12%2C104%2C55%2C56%2C51%2C10%2C34%2C77%5D%5D%2Cnull%2C%5C%22%token%%5C%22%5D%5D%22%2Cnull%2C%22generic%22%5D%5D%5D';
 
+const SPECIAL_COUNTRIES = ['kr', 'hk', 'tw'];
```
facundoolano, line 11: Can you use "korea" instead of "special"? I know it's not only Korea but still special is too vague to know when reading the code

facundoolano: Maybe `korea_store`

facundoolano:
Closing due to inactivity and conflict.

Issue #533, opened 2022-04-15
List function is returning an empty array


ozrendev:
* Operating System: Linux
* Node version: v17.7.1
* google-play-scraper version: github latest

### Description: 

`CLUSTER_NAMES` mapping needs to be updated for `TOP_FREE, TOP_PAID, GROSSING` otherwise it returns an empty array and cannot create `clusterUrl`

### Example code:
```js
gplay.list({
    category: gplay.category.COMICS,
    collection: gplay.collection.TOP_FREE,
    num: 2
  })
  .then(console.log, console.log);
```

### Error message:

`clusterUrl` returns empty array

facundoolano:
can you check if this still happens with the latest version?

facundoolano:
Issue automatically closed. Please create a new one if the issue persists

PR #534, opened 2022-04-15
Fix list mapping collection #533

ozrendev:


facundoolano:
Can you add a test to check for those collections? (so it fails next time this happens)

ozrendev:
Hey, I added a check inside the function to throw an error if it returns empty. 

Is this what you mean or within the list of unit tests?

facundoolano:
I mean in the unit tests

ozrendev:
When I looked further, it seems that the mappings will change depending on if you make a request with or without a category. I've made a fix for this and added tests if you could review it and let me know if this is the correct way to do this.

lib/list.js
```diff
@@ -82,17 +82,17 @@ const COLLECTIONS_MAPPING = ['ds:3', 0, 1];
...
-  GROSSING: 'top grossing apps',
-  TOP_FREE_GAMES: 'top free games',
-  TOP_PAID_GAMES: 'top paid games',
-  TOP_GROSSING_GAMES: 'top grossing games',
-  TRENDING: 'trending',
-  NEW_FREE: 'top new free android apps',
-  NEW_PAID: 'top new paid android apps',
-  NEW_FREE_GAMES: 'top new free games',
-  NEW_PAID_GAMES: 'top new paid games'
+  TOP_FREE: ['top free', 'top free apps'],
```
facundoolano, line 85: can you explain this change better, from the perspective of the user of this library? why do we need TOP_FREE to include these two cases? shouldn't it also include 'top free games' for that matter?

facundoolano:
Closing as other changes have been merged in this area.

PR #547, opened 2022-05-29
Fixed mapping for app listing

weizenberg:


facundoolano:
What's the difference with #545?

weizenberg:
It's missed the mapping for the 'search' method. I pulled from #545 and fixed the specific mapping.

Alexays:
@weizenberg it misses the fact of handling the enriched results, like facebook first result here 
e.g. https://play.google.com/store/search?q=facebook&c=apps

ninjachen:
Hello @weizenberg , I used your PR by 
`    "google-play-scraper": "git+https://github.com/facundoolano/google-play-scraper#pull/547/head"
`

but I still can not use the list function.  These are the steps I use
```
npm ci --include=dev
npx functions-framework --target=list
open a Chrome to http://localhost:8080/scraper-list?collection=topselling_free&country=us&from=0&limit=150
```

I failed with an error called 
```
Error: Invalid collection topselling_new_free
    at validate (/Users/ninja/src/lab/scraper/node_modules/google-play-scraper/lib/list.js:43:11)
    at /Users/ninja/src/lab/scraper/node_modules/google-play-scraper/lib/list.js:14:5
    at new Promise (<anonymous>)
    at Object.list (/Users/ninja/src/lab/scraper/node_modules/google-play-scraper/lib/list.js:13:10)
    at exports.list (/Users/ninja/src/lab/scraper/index.js:31:15)
    at /Users/ninja/src/lab/scraper/node_modules/@google-cloud/functions-framework/build/src/invoker.js:98:17
    at processTicksAndRejections (node:internal/process/task_queues:76:11)
```

Could you help on this list function?

ChayaGel:
It works me for search,
but some of the searches returns [], and I found why:
in the new UI, when google play find 1 spesific app that very suitable to the search term,
they show this app big and then the other results as "more results".
 (exapmle: https://play.google.com/store/search?q=waze&c=apps&hl=en&gl=us vs https://play.google.com/store/search?q=bank&c=apps&hl=en&gl=us)
**Such a situation must also be mapped.**
now it returns [].

I dont know how to fix the mapping, but at least I found the point.
can someone do it?

adtimizer:
@weizenberg Thanks for the PR. i'm trying to work with it but when the package name is incorrect it fails with a nasty error 

(node:33497) UnhandledPromiseRejectionWarning: HTTPError: Response code 404 (Not Found)
    at Request.request.once (/Users/orenbaranes/Adtimizer/Repo/gplay/node_modules/got/dist/source/as-promise/index.js:119:42)
    at process._tickCallback (internal/process/next_tick.js:68:7)
(node:33497) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:33497) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.

in the past it was returning an error of app not found 404.
can you help figure out what went wrong?

andrew-curry:
I used this pr but the gplay.list() returns [], anyone know how to deal with it?

ChayaGel:
This new mapping is excellent,
But it skips the "most appropriate" option for the search result ...!
We would expect that in case there is a particularly suitable one app, the function will return it first in the array and only after that will be added "more results"

thanks

weizenberg:
> This new mapping is excellent, But it skips the "most appropriate" option for the search result ...! We would expect that in case there is a particularly suitable one app, the function will return it first in the array and only after that will be added "more results"
> 
> thanks

Thank you, I added a check for the 'root' app type and added appropriate mapping. Check the recent commit.



> gplay.list() method still returns an empty array

As I mentioned before, this fix applies only to the **search** method. You're more than welcome to add the same mapping function to other methods as well.

facundoolano:
Closing this as it should be fixed by #557

KaKi87:
Hello,

Search works again, thanks !

However, why am I getting much less results than before ?
For example, `search({ term: 'web browser', num: 250 })` outputs 10 items.

Thanks

Issue #563, opened 2022-06-26
V9 has bugs


ChayaGel:
Fails "Cannot read property 'map' of undefined" on some cases - problems about new UI that are not fixed:
1.Countries issue seems still not work (I tried with "be")
(https://github.com/facundoolano/google-play-scraper/issues/551#issuecomment-1142043251 - this is the principle of solution, but have to come into the new version)
2. mapping result page with main-app not work, (try "waze" vs "waze_", only "waze_" worked properly) I saw that the code of v9 **is trying** to fix this, but something there is probably wrong.

knirpsfass:
First issue is on me.
I've analyzed it further once again in
https://github.com/facundoolano/google-play-scraper/issues/551#issuecomment-1167466402

ChayaGel:
works

Issue #551, opened 2022-05-31
Problem with countries in search


ChayaGel:
after all fixes in mapping from the "pull requests",
my search back to work.

but there are some strange problems:
search with country "be" (belgium) returns empty array- [], (maybe another countries)

knirpsfass:
I've debugged it a little bit. Seems like european? countries do have an additional
"About the result" information which shifts the index of the apps array by 1.
<img width="2212" alt="Screenshot 2022-05-31 at 13 56 16" src="https://user-images.githubusercontent.com/42298776/171167531-cf981ac4-17e3-4a97-992d-82b3443198a4.png">
 So the mapping for the apps would be 
```apps: ['ds:4', 0, 1, 1, 22, 0]```
instead of
```apps: ['ds:4', 0, 1, 0, 22, 0]```

example request for still working request: 
https://play.google.com/store/search?q=test&c=apps&hl=en&gl=US

example request for not working request:
https://play.google.com/store/search?q=test&c=apps&hl=en&gl=DE

facundoolano:
can you check if this still happens with the latest version?

knirpsfass:
works with 9.0.0

knirpsfass:
I fucked up really hard on the test. Really sorry for that. Issue is still present but now 
` "message": "Cannot read properties of undefined (reading 'map')"` is returned

PR #565, opened 2022-06-27
Fix country specific search

knirpsfass:
In this PR i fixed the country search since the new ui for the playstore introduced a **About the results** button for european countries.

Also i've changed the **more results** logic to no longer use a non localized string and isntead check if the value on a specific path is a string

For further information about the issue check: https://github.com/facundoolano/google-play-scraper/issues/551 and  #563

### DIFF
--- a/lib/search.js
+++ b/lib/search.js
@@ -101,7 +101,7 @@ async function processFirstPage (html, opts, savedApps, mappings) {
   if (sections.length === 1) {
     moreResultsSection = sections[0];
   } else {
-    moreResultsSection = sections.filter(section => R.is(String, R.path(SECTIONS_MAPPING.title, section)))[0];
+    moreResultsSection = sections.filter(section => isMoreSection(section))[0];
     if (moreResultsSection) {
       mainApp = scriptData.extractor(mainAppMapping)(R.path(mappings.app, html));
     } else if (R.path(SECTIONS_MAPPING.noResults, sections[0])) {
@@ -121,6 +121,11 @@ async function processFirstPage (html, opts, savedApps, mappings) {
   const token = R.path(SECTIONS_MAPPING.token, moreResultsSection);
 
   return checkFinished(opts, [...savedApps, ...apps], token);
+}
+
+function isMoreSection(section) {
+  const sectionTitle =  R.path(SECTIONS_MAPPING.title, section);
+  return R.is(String, sectionTitle) && !R.isEmpty(sectionTitle);
 }
 
 const INITIAL_MAPPINGS = {
--- a/lib/search.js
+++ b/lib/search.js
@@ -91,7 +91,7 @@ async function processFirstPage (html, opts, savedApps, mappings) {
   };
 
   const hasAboutTheseResults = R.is(String, R.path(mappings.aboutResultsTitle, html));
-  if(hasAboutTheseResults) {
+  if (hasAboutTheseResults) {
     R.path(mappings.aboutResultSection, html).shift();
   }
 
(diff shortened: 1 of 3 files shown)


facundoolano (changes requested):
Can you add a unit test to cover this case?

lib/search.js
```diff
@@ -118,9 +123,16 @@ async function processFirstPage (html, opts, savedApps, mappings) {
...
+  const sectionTitle = R.path(SECTIONS_MAPPING.title, section);
+  return R.is(String, sectionTitle) && !R.isEmpty(sectionTitle);
+}
+
 const INITIAL_MAPPINGS = {
   app: ['ds:4', 0, 1, 0, 23],
-  sections: ['ds:4', 0, 1]
+  sections: ['ds:4', 0, 1],
+  aboutResultSection: ['ds:4', 0, 1],
+  aboutResultsTitle: ['ds:4', 0, 1, 0, 31, 0]
```
mmoksh, line 135: It looks like `aboutResultsTitle` has the same mapping as `sections`. Why do you have to duplicate the element?

lib/search.js
```diff
@@ -118,9 +123,16 @@ async function processFirstPage (html, opts, savedApps, mappings) {
   return checkFinished(opts, [...savedApps, ...apps], token);
 }
 
+function isMoreSection (section) {
+  const sectionTitle = R.path(SECTIONS_MAPPING.title, section);
+  return R.is(String, sectionTitle) && !R.isEmpty(sectionTitle);
```
mmoksh, line 128: The first non-empty title does not guarantee it's the more_results section. Try a random query like `lajsdfkljew`. The first section should have a title: `No results for lajsdfkljew`. I think this case will break the script.

lib/search.js
mmoksh: Also, the `aboutResultsTitle` should be part of a SECTION mapping, not the initial mappings. It's something specific to one section.

lib/search.js
knirpsfass: This would return
```
{
    "message": "Keine Ergebnisse für <b>lajsdfkljew</b>"
}
```

lib/search.js
knirpsfass: as suggested its now using the section mapping + the result title mapping was moved to section specific mapping

knirpsfass:
> Can you add a unit test to cover this case?

i'll add tests asap

mmoksh:
Would you please add the following tests and make sure the code change doesn't break them?

1. query = "netflix"
=> expect(results[0] = "com.netflix.mediaclient")
=> expect("com.netflix.ninja" in results)
=> expect("?id=com.netflix.NGP.StrangerThings" in results)

2. query = "gmail"
=> expect(results[0] = "com.netflix.mediaclient")
=> expect(X in results) >> basically, make sure that the script reads the "more results" section, not the "similar apps" section

3. query = "netflix", hl=de, gl=de
=> expect(results[0] = "com.netflix.mediaclient")
=> expect("com.netflix.ninja" in results)
=> expect("?id=com.netflix.NGP.StrangerThings" in results)

4. query = "games"
=> expect("com.kiloo.subwaysurf" in results)

5. query = "kfdsjlkdsjifoewdsk"
=> expect empty set (currently it throws an error - we can change this behaviour)

knirpsfass:
@mmoksh am i mistaken or is the empty set not throwing an error on the master?

lib/search.js
knirpsfass: Okay never mind I'm kinda stupid😅. So yeah it still throws as it did before

ChayaGel:
UnhandledPromiseRejectionWarning: Error: No results for <b>cfgduyhgfuyjbgfuyhfuy</b>

knirpsfass:
@ChayaGel is this really an issue of this version?
With 9.0.0 the api no longer returns an empty array instead it throws - see https://github.com/facundoolano/google-play-scraper/issues/566#issue-1286959224

ChayaGel:
ok sorry

PR #570, opened 2022-07-02
Fixed similar method to support games

dbravobasecodeit:
Example: https://play.google.com/store/apps/details?id=co.yakand.agentaapuzzleindisguise&hl=en&gl=us only brings apps from the "More by..." section instead of the similar apps. This is because the scraper was looking for the "Similar apps" cluster and now it's called "Similar games" for games.

### DIFF
--- a/lib/similar.js
+++ b/lib/similar.js
@@ -71,7 +71,7 @@ function parseSimilarApps (similarObject, opts) {
   let similarAppsCluster = clusters.filter(cluster => {
     return R.path(CLUSTER_MAPPING.title, cluster) === SIMILAR_APPS ||
            R.path(CLUSTER_MAPPING.title, cluster) === SIMILAR_GAMES ||
-           clusters
+           clusters;
   });
   
   if (similarAppsCluster.length === 0) {
--- a/lib/similar.js
+++ b/lib/similar.js
@@ -68,12 +68,13 @@ function parseSimilarApps (similarObject, opts) {
   if (clusters.length === 0) {
     throw Error('Similar apps not found');
   }
+
   let similarAppsCluster = clusters.filter(cluster => {
     return R.path(CLUSTER_MAPPING.title, cluster) === SIMILAR_APPS ||
            R.path(CLUSTER_MAPPING.title, cluster) === SIMILAR_GAMES ||
            clusters;
   });
-  
+
   if (similarAppsCluster.length === 0) {
     similarAppsCluster = clusters;
   }
--- a/lib/similar.js
+++ b/lib/similar.js
@@ -61,13 +61,19 @@ const CLUSTER_MAPPING = {
 };
 
 const SIMILAR_APPS = 'Similar apps';
+const SIMILAR_GAMES = 'Similar games';
 
 function parseSimilarApps (similarObject, opts) {
   const clusters = R.path(INITIAL_MAPPINGS.clusters, similarObject);
   if (clusters.length === 0) {
     throw Error('Similar apps not found');
   }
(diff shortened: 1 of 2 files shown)


facundoolano (commented):
Can you add a unit test for this case?

mmoksh:
Could you please also change line 26 `hl: mergedOpts.lang` to `hl: 'en'`? Google Play displays the text `Similar apps/games` in different languages based on this parameter, and we only parse the cluster's url here, not the actual data.

dbravobasecodeit:
@facundoolano @mmoksh done

lib/similar.js
```diff
@@ -23,7 +23,7 @@ function similar (opts) {
 
     const qs = queryString.stringify({
       id: mergedOpts.appId,
-      hl: mergedOpts.lang,
+      hl: 'en',
```
facundoolano, line 26: why is this hard coded to 'en' now?

facundoolano: hrm I just saw @mmoksh comment. Are you positive that you don't get different results from different countries here? perhaps there's something different than the "Similar apps"/"Similar games" text that can be checked so it's not language dependant>?

mmoksh: @facundoolano I checked many apps with different countries/languages (including de, jp, ar), couldn't find any case where this fix won't work.

mmoksh: @facundoolano If we use a different language here, the `Similar apps/games` text would appear in a different language, and we cannot match the cluster title with the hard-coded text in English.

Now using `'en'` here would guarantee that we will get the correct cluster. The cluster url for different languages is different, but it returns the same list of apps. When the language parameter is appended to the cluster url in the next steps, it will get the apps' information in the desired language.

mmoksh: Example: https://play.google.com/store/apps/details?id=com.matchingham.impossible.date.tricky.riddles&hl=de

The cluster url for German is: [Link](https://play.google.com/store/apps/collection/cluster?gsr=SlNqGC9VNkJLS2U4QjEwY3NnNUxMWDNtdVE9PcICNgoyCi5jb20ubWF0Y2hpbmdoYW0uaW1wb3NzaWJsZS5kYXRlLnRyaWNreS5yaWRkbGVzEAcYCA%3D%3D:S:ANO1ljI2_Dk)

The cluster url for English is: [Link](https://play.google.com/store/apps/collection/cluster?gsr=SlNqGHpmVFY4NXJvQy80d0ZnZDFNeFh1akE9PcICNgoyCi5jb20ubWF0Y2hpbmdoYW0uaW1wb3NzaWJsZS5kYXRlLnRyaWNreS5yaWRkbGVzEAcYCA%3D%3D:S:ANO1ljLuO20)

When appending `hl=de` to the English cluster, it returns the same apps' info the first url gets: [Link](https://play.google.com/store/apps/collection/cluster?gsr=SlNqGHpmVFY4NXJvQy80d0ZnZDFNeFh1akE9PcICNgoyCi5jb20ubWF0Y2hpbmdoYW0uaW1wb3NzaWJsZS5kYXRlLnRyaWNreS5yaWRkbGVzEAcYCA%3D%3D:S:ANO1ljLuO20&hl=de)

mmoksh: @facundoolano I tried to find a different way to match the similar apps cluster, but couldn't find any. I even checked all properties for different clusters and couldn't find a different except the apps list, the title, and the random values that change with each request.

dbravobasecodeit:
Hi @facundoolano. Should I undo the change of the language? Please let me know!

lib/similar.js
dbravobasecodeit: @facundoolano  Let me know how you want to handle this.

facundoolano: We can leave it like this

PR #571, opened 2022-07-04
list: added num option for request to store / more than 50 apps returning in list method

Sidovsk:
After google's updates on play store, searches with the method list was returning only the top 50 applications, just like on the play store pages (example: https://play.google.com/store/apps/category/SHOPPING top free, top paid and grossing return only the first 45-50).
This PR allows list method returns more than 50 apps. Now the `num` option is passed to store, 
returning already in the request the number of apps needed . The maximum number of apps that is returning apparently is 660, probably limited by google in the search.


lib/list.js
```diff
@@ -113,10 +118,7 @@ async function parseCollectionApps (categoryObject, opts) {
     score: [0, 4, 1]
   };
   const appsPath = [0, 1, 0, 28, 0];
-  let processedApps = R.map(scriptData.extractor(appsMappings), R.path(appsPath, categoryObject));
-  if (processedApps.length > opts.num) {
```
mmoksh, line 117: Why did you remove this block?

PR #572, opened 2022-07-05
Fix app details mapping for asian countries

knirpsfass:
This PR fixes mapping of app details with the help of the service request id

DRAFT because tests are missing and   
mapping for
```
familyGenre: ['ds:5', 0, 12, 13, 1, 0],
familyGenreId: ['ds:5', 0, 12, 13, 1, 2],
```

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -33,7 +33,7 @@ function app (opts) {
     request(options, opts.throttle)
       .then(scriptData.parse)
     // comment next line to get raw data
-      .then(scriptData.extractor(MAPPINGS))
+      .then(scriptData.extractor(MAPPINGS, 'Ws7gDc'))
       .then(R.assoc('appId', opts.appId))
       .then(R.assoc('url', reqUrl))
       .then(resolve)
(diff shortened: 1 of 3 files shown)


knirpsfass:
Does anyone know what the `familyGenre` or `familyGenreId` is. I can't find an app that contains it

mmoksh:
Thanks for the fix! This may be a good test that makes the library's current version crash, and is fixed with this PR:
```
id: "com.squaretrade.personalcloud",
country: "au"
```

lib/app.js
```diff
@@ -33,110 +33,115 @@ function app (opts) {
     request(options, opts.throttle)
       .then(scriptData.parse)
     // comment next line to get raw data
-      .then(scriptData.extractor(MAPPINGS))
+      .then(scriptData.extractor(MAPPINGS, 'Ws7gDc'))
```
mmoksh, line 36: I checked a few apps and it looks like two sections have the same id. I'm not sure if the section that contains the details is the first one is a coincidence, or it will work for all cases. Just wondering if there could be another way to find the correct section. Maybe check the appId path for all sections and match the one that matches the input?

mmoksh:
```
        "ds:5": {
            "id": "Ws7gDc",
        },
        "ds:6": {
            "id": "Ws7gDc"
```

knirpsfass: I'm fine with the appId approach. I'm kinda busy right now so it might take some time for me to rewrite it.

mmoksh: @knirpsfass I just wrote it, would you please try it and see if it works better? All changes are in one file only.
I'm not sure what the familyGenre is, I would just leave it without changing.
```
    request(options, opts.throttle)
      .then(scriptData.parse)
    // comment next line to get raw data
      .then(appObject => getDetailsCluster(appObject, opts.appId))
      .then(scriptData.extractor(MAPPINGS))
      .then(R.assoc('url', reqUrl))
      .then(resolve)
      .catch(reject);
  });
}

function getDetailsCluster(appObject, appId) {
  for (let cluster in appObject) {
    if (R.path(MAPPINGS.appId, appObject[cluster]) === appId) {
      return appObject[cluster];
    }
  }
  // this should never run:
  return appObject["ds:4"];
}

const MAPPINGS = {
  appId: [1, 2, 77, 0],
  title: [1, 2, 0, 0],

```

mmoksh: Or you can do it without creating a new fuction:

```
request(options, opts.throttle)
      .then(scriptData.parse)
      .then(R.values)
      .then(clusters => clusters.filter(cluster => R.path(MAPPINGS.appId, cluster) === opts.appId)[0])
      .then(scriptData.extractor(MAPPINGS))
      .then(R.assoc('url', reqUrl))
      .then(resolve)
      .catch(reject);
```

mmoksh: One issue with this is that it will miss information in different sections. Like `comments` I think there should be separate mappings object for such fields.

facundoolano:
What's the status of this PR?

knirpsfass:
> What's the status of this PR?

Maybe I'll be able to implement the suggested changes this weekend otherwise upcoming weekend

knirpsfass:
closing this since meanwhile #583 fixed the issue

Issue #576, opened 2022-08-07
Similar is not working


josepius-clemson:
* Operating System: Windows 10
* Node version: v16.13.2
* google-play-scraper version: 9.0.2

### Description: 
Similar function is returning the following error:

### Example code:
```js
var gplay = require('google-play-scraper');
gplay.similar({appId: 'com.google.android.apps.translate'})
  .then(console.log, console.log);
```

### Error message:

```
Error: Error requesting Google Play:getaddrinfo ENOTFOUND play.google.comundefined&gl=us&hl=en
    at c:\..\node_modules\google-play-scraper\lib\utils\request.js:41:19
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  status: undefined
}
```

gianlucascoccia:
Is this working now? I get this error, no matter which app I try:

```
Error: Similar link not found
    at parseSimilarApps (/Users/gianlucascoccia/Desktop/store_scraper/node_modules/google-play-scraper/lib/parsers/parseSimilarApps.js:20:11)
    at /Users/gianlucascoccia/Desktop/store_scraper/node_modules/google-play-scraper/lib/similar.js:38:29
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
```

srikanthlogic:
Works for me, both in `9.1.1` and in `master`.

muzzyrost:
- Operating System: Ubuntu
- Node version: 16.16.0
- google-play-scraper version: 9.1.1

Example code:
```
var gplay = require('google-play-scraper');
gplay.similar({appId: 'com.google.android.apps.translate'}).then(console.log, console.log);
```
Error message:
```
TypeError: Cannot read properties of null (reading 'length')
at parseSimilarApps (/home/ub/node_modules/google-play-scraper/lib/similar.js:68:16)
at /node_modules/google-play-scraper/lib/similar.js:40:29
at processTicksAndRejections (node:internal/process/task_queues:96:5)

```

facundoolano:
Issue automatically closed. Please create a new one if the issue persists

PR #578, opened 2022-08-10
Fix similar script

mmoksh:
Fix issue: https://github.com/facundoolano/google-play-scraper/issues/576

Failing tests are related to the search script. Tests related to `similar` script are successful.

### DIFF
--- a/lib/similar.js
+++ b/lib/similar.js
@@ -50,7 +50,10 @@ function validateSimilarParameters (opts) {
 }
 
 const INITIAL_MAPPINGS = {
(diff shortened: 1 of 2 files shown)


muzzyrost:
Hi! Thanks for great fix! But I have little comment, when it parse some app from developer with few apps in account scraper takes field with _More by 'company name'_, and not _Similar apps_.

muzzyrost:
@mmoksh Hi! Thanks for great fix! But I have little comment, when it parse some app from developer with few apps in account scraper takes field with More by 'company name', and not Similar apps.

Issue #577, opened 2022-08-09
Search is not working again with 9.0.2 version


rsutariy:
* Operating System: MacOS 12.4 and Amazon Linux 2
* Node version: 16.13.1
* google-play-scraper version: 9.0.2

### Description: 
Search is working for some keywords and failing for others. May be meta data change again?
failed keywords examples : Wegmans, Facebook, Target, etc..

### Example code:
```
gplay.search({
    term: "facebook",
    num: 2
  }).then(console.log, console.log);
```

### Error message:
`Error TypeError: Cannot read properties of undefined (reading 'map') `

facundoolano:
Issue automatically closed. Please create a new one if the issue persists

PR #581, opened 2022-08-13
Fix search script

mmoksh:
Fix errors in search script: https://github.com/facundoolano/google-play-scraper/issues/577


mmoksh:
@facundoolano

PR #599, opened 2022-10-14
Replace querystring with URLSearchParams

pavi2410:


facundoolano:
any particular reason for this change?

pavi2410:
Deno / Cloudflare Workers support. Because `querystring` module doesn't exist outside Node.

facundoolano:
hrm I think if that's the case the right thing to do here would be to add a github check that this runs on deno, otherwise it could break again with further changes

pavi2410:
It won't break as it uses an standard web API and I tested this when working on https://github.com/pavi2410/PlayBadges

However, Deno or better, standard web API compatibility would be awesome.

facundoolano:
Closing this as it doesn't make sense to make an isolated change for deno support without adding the rest of the infra to support it.

PR #600, opened 2022-10-16
Fixing search for the A/B testing

CyrilCermak:
Credit goes to @Djip, I only tested the change in my app and it seems to work perfectly fine!


facundoolano:
the lint check was failing on this one

PR #605, opened 2022-11-08
Fix README links

Mahrkeenerh:
Change links using `dev` branch to `main` branch
Update line references for `Category` and `Collection`

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,8 +112,8 @@ Results:
 ### list
 Retrieve a list of applications from one of the collections at Google Play. Options:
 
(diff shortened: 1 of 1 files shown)


facundoolano:
Thanks. It would be best to use specific commits, though, so the links stay relevant for readers even if the files change. You can get them by pressing 'y' while looking at the file in GitHub, for example:

https://github.com/facundoolano/google-play-scraper/blob/e7d9dec56d9bf0df6a115be9ebed5e91534791d1/lib/constants.js#L67

Mahrkeenerh:
but if the files change, they could be not relevant anymore, no? If they decide to add a new category for example, and the link still points to the old version ...

But I guess that's less likely to happen, I'll update them soon

facundoolano:
right, neither option is ideal but I rather have the link point to something understandable and then if they want to get the latest version they need to change explicitly to main (or HEAD). Otherwise some other thing may change in the file and the link points to some weird moved position.

thanks!

Issue #404, opened 2020-05-23
Suggested feature: allow "machine default" country / lang through an options parameter


icarcal:
### Description: 
As suggested [here](https://github.com/facundoolano/google-play-scraper/issues/400#issuecomment-632824987), remove the `hl` and `gl` from request url in a method call through an option parameter

### Why?
[Some apps are not published in the us](https://github.com/facundoolano/google-play-scraper/issues/400#issuecomment-632466973), but are accessible through common gplay url

### Requirements
- Default behaviour in methods should not change (all methods must still use `us` as country and `en` as language), if this option is not present

- This option needs to be respected in [all method calls](https://github.com/facundoolano/google-play-scraper/issues/400#issuecomment-633087547)

mnikitas:
Any news to this request. It is **important** to have the option to remove hl and gl from request url in order to be able to find apps only available to a specific country without requiring from user to specify a country.

This is a very easy fix in code as we've seen.

Can you please arrange this change ?

Regards,
Mike

facundoolano:
I'd be ok with such an option, such as passing explicitly country: undefined / lang: undefined in the options.

I don't have time planned for making changes to the library, but I can review and merge if someone sends the PR.

facundoolano:
Issue automatically closed. Please create a new one if the issue persists

PR #627, opened 2023-04-05
feat: allow remove hl gl if -1 in them

riderx:
by passing -1 hl and gl are removed in request as requested in #404
I choosed -1 instead of undefined or false to be sure when argument is ommited nothing is changed.

fix the issue with forced 'hl':'en' for similar.
fix the issue where gl was missing in datasafety

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -20,8 +20,8 @@ function app (opts) {
 
     const qs = queryString.stringify({
       id: opts.appId,
-      ...(opts.lang === -1 ? {} : {hl: opts.lang}),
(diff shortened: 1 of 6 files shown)


facundoolano:
-1 is not a good value for this, it would be better to introduce a [constant](https://github.com/facundoolano/google-play-scraper/blob/HEAD/lib/constants.js) for this, with a readable name and string value, eg. `LOCAL_DEFAUILT_LANGUAGE` and `LOCAL_DEFAULT_COUNTRY`

facundoolano:
closing due to inactivity

PR #634, opened 2023-05-12
Partial fix for unit tests

tpwatson:
This mostly fixes the unit tests with some very minor updates.

The last two tests are in relation to suggested search results, and the fix for that would be much more demanding. I think if I continue trying to tackle it I would make it a seperate PR because I would need to some heavy modifications to the search.js file.


tpwatson:
The first test "should fetch valid application data" was failing. 

After looking into why, its appears that Panda Pop's preview video is currently not loading (as of the past 10 minutes!) for hl=en&gl=US..

PR #636, opened 2023-05-16
Fixing the unit tests.

tpwatson:
Hi, @facundoolano I've got all the tests passing now.

The biggest issue was with search.js, everything else just needed some updated assert values.

With search.js I had to create a way to detect which mappings would be needed from the search results.

There were two new mapping sets which only applied to:
1) Suggested Results for EU countries
2) Suggested Results for US

The previous mappings still worked for non-Suggested Results responses, but would not work in the final two test cases for the unit tests.

So basically I just create the two new mappings sets, and then provided a way for search.js to know when to use each one.

This is my first Pull Request, so let me know if what I'm trying to say is not clear, or what else I need to do!

Thanks!

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -140,7 +140,7 @@ const MAPPINGS = {
   },
   recentChanges: ['ds:5', 1, 2, 144, 1, 1],
   comments: {
-    path: ['ds:9', 0],
+    path: ['ds:8', 0],
     isArray: true,
     fun: helper.extractComments
   }
--- a/lib/app.js
+++ b/lib/app.js
@@ -40,7 +40,13 @@ function app (opts) {
       .catch(reject);
   });
 }
-
+/*
+  * If you need help checking and fixing mappings do the following:
+  1 If you cannot use devtools to get the entire object to copy, run gplay.search and write the HTML response to a file
+  2 Copy the javascript object from the AF_initDataCallback and paste it into https://www.convertsimple.com/convert-javascript-to-json/ (copy starting with outermost {} )
+  3 Next copy that output which is now in JSON format and paste it into https://jsonpathfinder.com/
+  4 You can now select each field until you see the exact path to the data you want
+*/
 const MAPPINGS = {
   title: ['ds:5', 1, 2, 0, 0],
   description: {
--- a/lib/search.js
+++ b/lib/search.js
@@ -128,8 +128,8 @@ function isMoreSection (section) {
 function removeUnneededSections (html, mappings) {
   removeSectionsIfPathValueOfType(html, SECTIONS_MAPPING.aboutResultsTitle, String);
   // if the search function does no longer throw when no result was returned
-  // mapping SECTIONS_MAPPING.suggestedResultDescritpion can be replaced with SECTIONS_MAPPING.noResult mapping
-  removeSectionsIfPathValueOfType(html, [...mappings.sections, 0, ...SECTIONS_MAPPING.suggestedResultDescritpion], String);
+  // mapping SECTIONS_MAPPING.suggestedResultDescription can be replaced with SECTIONS_MAPPING.noResult mapping
+  removeSectionsIfPathValueOfType(html, [...mappings.sections, 0, ...SECTIONS_MAPPING.suggestedResultDescription], String);
   removeSectionsIfPathValueOfType(html, SECTIONS_MAPPING.aboutResultsTitle, String);
 }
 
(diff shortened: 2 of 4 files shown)


tpwatson:
Well now the similar apps don't want to see apps from other developers. Working on fixing that.

lib/app.js
```diff
@@ -40,7 +40,13 @@ function app (opts) {
       .catch(reject);
   });
 }
-
+/*
```
facundoolano, line 43: this is out of scope for this PR, but this sort of info should perhaps be in other doc, and we could eventually add some sort of script to do that work instead of relying on those websites

lib/search.js
```diff
@@ -34,11 +45,62 @@ function initialRequest (opts) {
...
+let SUGGESTED_APPS_MAPPING = [];
+let SUGGESTED_APPS_EU_MAPPING = [];
 
 async function processFirstPage (html, opts, savedApps, mappings) {
   if (R.is(String, html)) {
     html = scriptData.parse(html);
   }
+  // For US country code searches, if suggested results are given the mappings will be different
+  // change these mappings to fix any suggest search issues related to US country code
+  if (opts.country === 'US' && typeof R.path(['ds:4', 0, 1, 0, 25, 0, 1, 1, 1], html) === 'string') {
```
facundoolano, line 72: can you update this to make it independent of specific country? we don't know for what subset of countries the different layout is returned, it could change in the future even if now it's us only (or US could be changed to match the rest). If there are two different cases ideally the code should figure out which one to use by inspecting the data.

lib/search.js
```diff
@@ -34,11 +45,62 @@ function initialRequest (opts) {
 function extaractDeveloperId (link) {
   return link.split('?id=')[1];
 }
+// Mappings used for normal search results (no misspellings aka no suggested results)
```
facundoolano, line 48: would it make this simpler if we dropped support for suggested/mispelling results? can you show how it looks on the ui and make a case for supporting this as part of the results here?
(there's a separate method for search suggestions not sure if they are the same sort of thing)

test/lib.search.js
```diff
@@ -64,19 +64,20 @@ describe('Search method', () => {
...
 
   describe('more results mapping', () => {
-    it('schould return few netflix apps', () => {
+    it('should return few netflix apps', () => {
       return gplay.search({ term: 'netflix' })
         .then((apps) => {
           assert.equal(apps[0].appId, 'com.netflix.mediaclient');
-          assertIdsInArray(apps, 'com.netflix.ninja', 'com.netflix.NGP.StrangerThings');
+          // assert there are at least two apps with '.netflix' in the id
+          assertIdsInArray(apps, 'com.netflix.ninja', 'com.netflix.NGP.Poinpy' || 'com.netflix.NGP.SpongeBobKrustyCookOff' || 'com.netflix.NGP.StrangerThings' || 'com.netflix.NGP.Oxenfree');
```
facundoolano, line 72: how does this OR (||) work here? I think it's best to not use assertIdsInArray here and assert manually for the condition described in the comment

lib/search.js
```diff
@@ -185,7 +243,8 @@ function search (appData, opts) {
     opts = {
       term: encodeURIComponent(opts.term),
       lang: opts.lang || 'en',
-      country: opts.country || 'us',
+      // make opts.country uppercase to avoid case issues
```
facundoolano, line 246: why uppercase?

lib/search.js
```diff
@@ -34,11 +45,62 @@ function initialRequest (opts) {
...
+      token: [25, 1, 3, 1],
+      noResults: [25, 0, 0, 0, 1],
+      suggestedResultDescription: [25, 0, 0, 0, 1],
+      aboutResultsTitle: ['ds:4', 0, 1, 0, 31, 0]
+    };
+  }
+
+  // For EU country codes, if suggest results are given, the mappings will be different again
+  // changet these mappings to fix any suggest search issues related to EU country codes
+  const EU_CODES = ['AT', 'BE', 'BG', 'CY', 'CZ', 'DE', 'DK', 'EE', 'ES', 'FI', 'FR', 'GR', 'HR', 'HU', 'IE', 'IT', 'LT', 'LU', 'LV', 'MT', 'NL', 'PL', 'PT', 'RO', 'SE', 'SI', 'SK'];
```
facundoolano, line 89: same as before we shouldn't have to track a list of EU countries nor assume that EU will be a separate case or that all countries within EU will have the same layout.

facundoolano:
closing this one as a tests have been mostly fixed by #645

PR #646, opened 2023-08-03
Esm Migration + bump libraries

jbigman:
### Dependencies
Bump got from 11.8.6 to 13.0.0
Bump mocha from 9.1.4 to 10.2.0
And minor updates.

### ESM migration
Not that much to say.


facundoolano:
Thanks for this contribution, but this PR is kind of big and it's fixing changing several separate things at once. Can you make separate PRs?

For example, the title says it fixes the tests but I see the build is still failing. The ideal would be to get a PR to fix the tests and once that is working get a separate one for the esm so we can see that nothing breaks, and separate ones for library bumps, etc.

facundoolano:
Well, my suggestion is to work at one of the problems at a time. You can start by the simplest one if you prefer, to make it easier to get things merged before getting into the complicated parts. For example, the mapping updates should be simple to review on their own, you can branch from the current main and apply only those changes, then wait for that to get merged before moving to the next PR.

jbigman:
#649 Should definitly fix checks.
#648 Add new properties

This current PR fixes Esm migration and bumps. Do you want to fix imports before bumps ?

facundoolano:
Please pull from main to get a clean diff. Also, yes it would be better to separate the version bumps so this is as small as possible for review

lib/app.js
```diff
@@ -1,11 +1,9 @@
-'use strict';
-
-const R = require('ramda');
-const queryString = require('querystring');
-const request = require('./utils/request');
-const scriptData = require('./utils/scriptData');
-const { BASE_URL } = require('./constants');
-const helper = require('./utils/mappingHelpers');
+import { assoc, path } from 'ramda';
```
facundoolano, line 1: I noticed that you replaced a lit of module imports eg R for direct imports of module functions. Please don't do that, as we lose the information of where each function is coming from, making the code less readable

facundoolano:
Closing this one as ESM migration is happening in #651

Issue #632, opened 2023-04-12
Migrate to ESM


facundoolano:
Replace the old commonjs requires with ESM imports. This is preventing an update of the latest version of the got lib at the moment, and I'm assuming that in general it's the preferred default nowadays.

tpwatson:
I got the migration done just now.
After fixing all the errors one by one it appears to be functioning 100%
The fact that every unit test gave a result without an error tells me the migration is complete.
![Untitled](https://github.com/facundoolano/google-play-scraper/assets/8966870/b73aa366-291b-4fed-9e47-b7302b44f0c2)

I create a seperate branch on my fork of this repo. I'm just trying to figure out how to push to that so I can submit a PR for this.

tpwatson:
https://github.com/facundoolano/google-play-scraper/pull/637

facundoolano:
closed by #651

PR #651, opened 2023-08-04
Esm migration

jbigman:


lib/categories.js
```diff
@@ -1,8 +1,6 @@
-'use strict';
-
-const request = require('./utils/request');
-const cheerio = require('cheerio');
-const { BASE_URL } = require('./constants');
+import request from './utils/request.js';
+import { load } from 'cheerio';
```
facundoolano, line 2: you should be able to do it like this, as per [the documentation](https://github.com/cheeriojs/cheerio#loading):

```
import * as cheerio from 'cheerio';

// ...

.then(cheerio.load)
```

lib/app.js
```diff
@@ -40,7 +38,13 @@ function app (opts) {
       .catch(reject);
   });
 }
-
+/*
```
facundoolano, line 41: we shouldn't have instructions pointing to 3rd party online services. 
also this is not the best place to put this kind of documentation. 
It is out of the scope of this PR, but it's preferrable to add a separate markdown file to the repo and link it from here. And if this sort of operation is necessary to fixing mappings, these instructions should be turned into a script with native node.js implementation of the operations (e.g. why use a 3rd party website to output JSON?)

facundoolano: so remove this comment and you can put the content into an issue to plan for adding the necessary support doc/scripts

facundoolano (changes requested):
Left a couple of change requests.

Also, you should update the README, since most examples are still using `require` (that won't be supported anymore right?)

README.md
```diff
@@ -1,5 +1,5 @@
 # google-play-scraper [![workflow](https://github.com/facundoolano/google-play-scraper/actions/workflows/tests.yml/badge.svg)](https://github.com/facundoolano/google-play-scraper/actions/workflows/tests.yml)
-Node.js module to scrape application data from the Google Play store.
+Node.js [ES](https://262.ecma-international.org/6.0/) module to scrape application data from the Google Play store.
```
facundoolano, line 2:
```suggestion
Node.js module to scrape application data from the Google Play store.
```

facundoolano (approved):
LGTM, thanks!!

facundoolano:
this one closes #632

facundoolano:
@jbigman just merged, feel free to send a separate PR with library bumps and I'll publish a major version after that.

PR #654, opened 2023-08-11
Replace throttle implementation by external library

jbigman:


jbigman:
Don't mind about the third commit description, i missed a copy paste. it contains eslint adjustments.

lib/utils/processPages.js
```diff
@@ -26,18 +28,24 @@ async function processPages (html, opts, savedApps, mappings) {
...
+    appId: app.appId,
+    lang: opts.lang,
+    country: opts.country,
+    cache: opts.cache,
+    throttle: opts.throttle,
+    requestOptions: opts.requestOptions
+  });
+
+  if (opts.throttle !== undefined) {
+    const throttle = pThrottle({
```
facundoolano, line 41: why did you need to add this throttle setup? doesn't the setup in the internal request module apply here as well by just passing the throttle option?

jbigman:
I wrote this implementation

```js
const req = (resolve, reject) => requestLib(opts)
    .then((response) => resolve(response.body))
    .catch((error) => reject(error));

  let promiseToResolve = (resolve, reject) => req(resolve, reject);

  if (throttleParams !== undefined) {
    const throttle = pThrottle({
      limit: throttleParams.limit ? throttleParams.limit : 1000,
      interval: throttleParams.interval ? throttleParams.interval : 0
    });
    promiseToResolve = (resolve, reject) => throttle(req(resolve, reject));
  }
  return new Promise(promiseToResolve);
```
  
  But we could use the throttling method everytime with default values set to  {limit: 1000, interval : 0}  
  
```js
  const req = (resolve, reject) => requestLib(opts)
    .then((response) => resolve(response.body))
    .catch((error) => reject(error));

  const throttle = pThrottle({
    limit: throttleParams && throttleParams.limit ? throttleParams.limit : 1000,
    interval: throttleParams && throttleParams.interval ? throttleParams.interval : 0
  });
  return new Promise((resolve, reject) => throttle(req(resolve, reject)));
```

lib/utils/processPages.js
jbigman: I thougth the same, but test fails if i don't edit processpage when it batch request fulldetails

jbigman:
Both the `limit `and `interval `options must be specified in p-throttle usage. 

That's why i force interval or limit with default values if one is missing

facundoolano:
so a couple of things:

1. by default throttling should be off. we don't want to silently introduce a limit unless the client explicitly tell us to do so
2. it's ok to set a default to the missing argument if the other is specified.
3. see the [related documentation](https://github.com/facundoolano/google-play-scraper#throttling) on this repo. You should make sure we continue to support the same behavior, and update accordingly if there are new options or things to be considered (but it's best not to introduce a change in behavior or interface if not necessary). In particular, I think it's fine to continue to assume the interval is always seconds, and just take an integer as the throttle value, which will be assumed to be the request limit by second.
4. I need to review more closely that duplicated setting you have for the fullDetail, I want to understand why it's necessary

lib/utils/processPages.js
jbigman: My bad, i reverted changes on this files, it now works perfectly, i'm on it.

jbigman:
There is no singleton implemented in the library, throttling is done only on batch requests. requested api outside are not handled by the p-throttle implementation.

I've tested localy, all tests are ok if I use this in request.js
```js
function doRequest (opts) {
  return new Promise((resolve, reject) => requestLib(opts)
    .then((response) => resolve(response.body))
    .catch((error) => reject(error)));
}
```
and 
```js
async function processFullDetailApps (apps, opts) {
  const getAppDetails = (app) => appDetails({
    appId: app.appId,
    lang: opts.lang,
    country: opts.country,
    cache: opts.cache,
    throttle: opts.throttle,
    requestOptions: opts.requestOptions
  });

  if (opts.throttle !== undefined) {
    const throttle = pThrottle({
      limit: opts.throttle.limit ? opts.throttle.limit : 1000,
      interval: opts.throttle.interval ? opts.throttle.interval : 0
    });
    return Promise.all(apps.map(await throttle(getAppDetails)));
  } else {
    return Promise.all(apps.map(getAppDetails));
  }
}

```

jbigman:
in fact, p-throttle doesnt queue requests, i'll try to use [throttledQueue ](https://github.com/shaunpersad/throttled-queue) instead

facundoolano:
closing due to inactivity, but feel free to send a new one.

PR #669, opened 2023-11-08
Safer screenshots mapping

Aliaksandr-Kasko-JazzTeam:
Do not crush when screenshots is undefined

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -136,7 +136,7 @@ const MAPPINGS = {
   screenshots: {
     path: ['ds:5', 1, 2, 78, 0],
     fun: (screenshots) => {
-      if (screenshots === null) return [];
+      if (!screenshots?.length) return [];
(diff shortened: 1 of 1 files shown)


lib/app.js
```diff
@@ -136,7 +136,7 @@ const MAPPINGS = {
   screenshots: {
     path: ['ds:5', 1, 2, 78, 0],
     fun: (screenshots) => {
-      if (screenshots === null) return [];
+      if (!screenshots?.length) return [];
```
facundoolano, line 139: is this different from?
```suggestion
      if (!screenshots) return [];
```

Aliaksandr-Kasko-JazzTeam: Yes, it's different. If you remove `?.length`, then this if goes falsy and code on the line 140 will be executed, but when screenshots is empty, then map returns an empty array.
So, if you leave part `?.length`, then this check can be a little quicker.

Issue #676, opened 2024-04-03
Variable number of reviews returned


gianlucascoccia:
The reviews methods seems to return a different number of reviews on each call.
I slightly modified the example in the documentation to reproduce the issue:

```
import gplay from "google-play-scraper";

gplay.reviews({
  appId: 'com.facebook.katana',
  num: 3000
}).then(function(apps){
  console.log(apps.data.length);
}, console.log);
```

When called multiple times the output is :

```
300
300
150
150
150
150
1200
```

This  used to work in the past, getting hundreds of thousands of reviews in a single call

sheldon0711:
Experience the same issue. Now reviews_all function only returns very limited reviews, instead of all reviews in the past.

gianlucascoccia:
In [this sister project](https://github.com/JoMingyu/google-play-scraper/) they have encountered the same issue.
[Here](https://github.com/JoMingyu/google-play-scraper/issues/209#issuecomment-2036302883) I provided  a workaround around the issue that appears to be working.
I would submit a PR to fix the issue but my JavaScript skills are fairly limited.

petskratt:
Looks like Google has implemented limit of 150 items per response, so paging is needed. And with paging comes need to provide `NID` cookie so loadbalancer sends your request to proper backend that knows about `nextPaginationToken` from pervious request (if not, then I rarely get more than 4-5 pages of 150 items).

To validate my theory I modified https://github.com/facundoolano/google-play-scraper/blob/main/lib/reviews.js#L154 adding `NID` value from my browser ... and got paging working:

```javascript
    headers: {
      'Content-Type': 'application/x-www-form-urlencoded;charset=UTF-8',
      'Cookie': 'NID=[cookie value from browser];'
    }
```

Real solution means implementing `cookieJar`, perhaps in https://github.com/facundoolano/google-play-scraper/blob/main/lib/utils/request.js

```javascript
import gplay from "google-play-scraper";

const fetchAllReviews = (appId, nextPaginationToken) => {
    return gplay.reviews({appId: appId, paginate: true,  nextPaginationToken: nextPaginationToken})
        .then (
            (res) => res.nextPaginationToken
                ? fetchAllReviews (appId, res.nextPaginationToken).then (fullData => [...res.data, ...fullData])
                : res.data
        )
}

fetchAllReviews ('com.some.thing', null)
    .then (console.log)
    .catch (console.warn)
```

PR #677, opened 2024-04-07
Fix variable number of reviews returned #676

petskratt:
According to issue #676 reviews (and presumably other listings) that require pagination are returning inconsistent results. Implementing cookieJar appears to fix that.

Not tested with other functions.

### DIFF
--- a/lib/utils/request.js
+++ b/lib/utils/request.js
@@ -29,7 +29,7 @@ function doRequest (opts, limit) {
   });
 }
 
-async function request(opts, limit) {
+async function request (opts, limit) {
   debug('Making request: %j', opts);
   try {
     const response = await doRequest(opts, limit);
--- a/lib/utils/request.js
+++ b/lib/utils/request.js
@@ -1,11 +1,16 @@
 import requestLib from 'got';
 import throttled from './throttle.js';
+import {CookieJar} from 'tough-cookie';
 import createDebug from 'debug';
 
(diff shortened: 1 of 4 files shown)


petskratt:
All fails appear to be apps that have been removed from store - `air.com.zinkia.playset` (app method), `air.tv.ingames.cubematch.free` and `com.skybornegames.battlepop` (permisssions method).

lib/utils/request.js
```diff
@@ -1,11 +1,16 @@
...
+import { CookieJar } from 'tough-cookie';
 import createDebug from 'debug';
 
+const cookieJar = new CookieJar();
 const debug = createDebug('google-play-scraper');
 
 function doRequest (opts, limit) {
   let req;
+
+  opts.cookieJar = cookieJar;
```
facundoolano, line 12:
```suggestion
  // cookies are necessary for pagination to work consistently across requests
  opts.cookieJar = cookieJar;
```

facundoolano:
this LGTM. perhaps you can extend the pagination related test to check that the same query returns the same results every time (ideally this would be failing in main and would be fixed by this change)

petskratt:
Good point - added test that utilises built-in paging, fails on main and works with cookies. As several other tests fail due to discontinued apps ... com.facebook.katana will hopefully stay up for some time and considering the number of reviews perhaps the set of 1500 most helpful ones does not change too often.

Issue #682, opened 2024-08-08
Comments Location Regressed


jonathansampson:
> This issue has regressed. @jbigman, comments are now in `ds:9` again. Perhaps the extractor should check both `ds:8` and `ds:9` (one being a fallback if the other doesn't contain comments). Thoughts?

_Originally posted by @jonathansampson in https://github.com/facundoolano/google-play-scraper/issues/603#issuecomment-2276305708_

PR #683, opened 2024-08-08
dynamically resolves comments path

jonathansampson:
Tests are failing in part due to an inability to resolve the proper comments path. This is a step towards resolving test issues.

fixes https://github.com/facundoolano/google-play-scraper/issues/682

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -160,7 +160,7 @@ const MAPPINGS = {
   },
   recentChanges: ['ds:5', 1, 2, 144, 1, 1],
   comments: {
-    path: ['ds:8', 0],
+    path: [],
     isArray: true,
     fun: helper.extractComments
   },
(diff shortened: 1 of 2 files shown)


lib/app.js
```diff
@@ -160,7 +186,7 @@ const MAPPINGS = {
   },
   recentChanges: ['ds:5', 1, 2, 144, 1, 1],
   comments: {
-    path: ['ds:8', 0],
+    path: [COMMENTS_PATH, 0],
     isArray: true,
     fun: helper.extractComments
```
facundoolano, line 191: there's already an extract comments helper here, and it seems to only be used by this module. I suggest that you move helper.extractComments over this module and extend it to do this selective logic instead of adding the new determineCommentsPath. 

(in general, it's preferable to use an extract `fun` to customize parsing and handle these special cases instead of adding more steps to the top level function chain)

facundoolano:
Looks good, thanks

Issue #687, opened 2024-08-09
Access short description from app listing


PierfrancescoSoffritti:
The library provides an API to fetch data about an app. The app detail object returned by this API contains the full description of the app, but not the short description. 

Is there a reason why the short description is not included? I can see it if I inspect the source code of the Play Store page.

PierfrancescoSoffritti:
Sure, every app has one [see screenshot from Google Play](https://github.com/user-attachments/assets/37bf1a43-5462-4de1-93d6-4b077d68321b).

It looks like they are only visible from Google Play's mobile UI, not from the desktop UI. [See screenshot from mobile UI](https://dl3.pushbulletusercontent.com/VsNemFfQhiRgHqNjV2xDJO1W45z33eUw/Screenshot_20240809-193119.png).

Somewhat annoyingly the source code of the web page only mentioned it as "description". You can search for "AI video editor with music!" in the source of [this page](https://play.google.com/store/apps/details?id=com.camerasideas.instashot).

jonathansampson:
Looks like the short description is just adjacent to the long description, so trivial to include. I've issued a PR to add the property and data, @PierfrancescoSoffritti. Nice suggestion :)

PierfrancescoSoffritti:
Nice, thanks for the very fast PR!

jonathansampson:
It looks like `summary` has the value you're seeking.

facundoolano:
@PierfrancescoSoffritti you are looking for the `summary` property. Closing this as it's already implemented.

PR #688, opened 2024-08-09
introduces support for shortDescription

jonathansampson:
addresses https://github.com/facundoolano/google-play-scraper/issues/687

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -49,6 +49,7 @@ const MAPPINGS = {
     path: ['ds:5', 1, 2],
     fun: helper.descriptionHtmlLocalized
   },
+  shortDescription: ['ds:5', 1, 2, 73, 0, 1],
   summary: ['ds:5', 1, 2, 73, 0, 1],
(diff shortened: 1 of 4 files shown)


PierfrancescoSoffritti:
Thanks for the prompt PR!

nit: you might wanna add a mention of the short description to [the example in the README](https://github.com/facundoolano/google-play-scraper?tab=readme-ov-file#app) as well :)

jonathansampson:
Good call, @PierfrancescoSoffritti.

lib/app.js
```diff
@@ -49,6 +49,7 @@ const MAPPINGS = {
     path: ['ds:5', 1, 2],
     fun: helper.descriptionHtmlLocalized
   },
+  shortDescription: ['ds:5', 1, 2, 73, 0, 1],
```
facundoolano, line 52: Isn't this just the same as the summary below?

jonathansampson: Whoa. It is. Sorry, @PierfrancescoSoffritti, it looks like `summary` has what you're seeking :)

PR #693, opened 2024-09-24
Added new about the developer section to the `app` method

renatosrounds:
Hello,

I wanted to add the new "About the developer" section which I think will soon be mandatory to all apps. (screenshot as reference)

I split it in its own object so its better organised. Let me know if you think that is okay.

Screenshot 2024-09-24 at 09 59 48

### DIFF
--- a/lib/app.js
+++ b/lib/app.js
@@ -110,6 +110,21 @@ const MAPPINGS = {
   developerEmail: ['ds:5', 1, 2, 69, 1, 0],
   developerWebsite: ['ds:5', 1, 2, 69, 0, 5, 2],
   developerAddress: ['ds:5', 1, 2, 69, 2, 0],
+  aboutDeveloper: {
+    path: ['ds:5', 1, 2, 69],
+    fun: (searchArray) => {
+      const name = R.path([4, 0], searchArray);
+      const email = R.path([4, 1, 0], searchArray);
+      const address = R.path([4, 2, 0], searchArray).replace(/\n/g, ', ');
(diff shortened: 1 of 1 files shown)


facundoolano:
Let's call it devInfo, and update the readme please and the tests accordingly, please.

facundoolano:
Also I don't think this should be a separate object (I mean it would be best to have all dev fields into a separate object, but we're already using separate developerId, developerEmail, etc, and I prefer not to introduce a breaking change).

You should use developerLegalName or developerCompany or something similar to avoid confusion, Likewise check whether the email in the about section is different from the email we already have and rename if necessary.

renatosrounds:
Ok I will remove them from the object. I have noticed in some apps that the email can be different between this new section and the support email in the about section.

renatosrounds:
@facundoolano On the app that is used on the README and the tests, I see that the `About the developer` section is not present yet. So in the README I put the values as undefined while in the tests I just added a new one instead of updating the validation function.
Do you have a better idea on how I can handle this? Should I change the app in the tests for example?

facundoolano:
Code looks good, the only thing is that for the test, you should try to choose an app that's less likely to be removed to reduce the chances of breaking the test down the line. Did you chose this developer for some particular reason?

renatosrounds:
No you are right. I will find an app that has the information and has less chance of being removed.

renatosrounds:
@facundoolano I updated the app with soundcloud which should be more solid as a test

PR #695, opened 2024-10-16
feat(permissions): add default country and update URL

Cristianoying:
Fix the situation that permissions cannot be queried without a country code in some cases.
- Add default country ('us') if not provided in opts
- Update batchexecute URL to include country (gl) parameter


facundoolano:
what's up with this one, is it ready to merge? (I see it's still in draft)

Cristianoying:
I was thinking of adding some test code but I've been so busy forgot to add it.😂 but I've tested it in my production environment.

PR #704, opened 2025-01-08
fix typescript error.

talrasha007:
Screenshot 2025-01-08 at 09 55 12

when import from typescript project, there's a error that will cause tsc build fail (strict mode), add default export will fix this issue.


talrasha007:
sorry, it seems #702 is a better.

PR #705, opened 2025-01-08
Fixed few tests

jbigman:
Tests : 
- should fetch valid application data
- should fetch valid application data for country: es
- should fetch valid application data for country: br

Where failing because of missing reviews.
=> updated mapping to also search in ds:10

Test 
- should fetch games from different developers

Where failing because only one developer was present in results.
=> Used another app Id


jbigman:
merge ?

PR #708, opened 2025-01-30
reviews sorting body structure fix

tugbargn:
Key Changes:
The original payload was URL-encoded (%5B, %22, etc.), which made it harder to read and debug. I decoded it into a more readable JSON-like structure.

I adjusted the structure of the array inside the payload to match the expected format for the Google Play API. 
Specifically:

The sort and numberOfReviewsPerRequest variables are now correctly placed in the array.

### DIFF
--- a/lib/reviews.js
+++ b/lib/reviews.js
@@ -80,7 +80,7 @@ function getBodyForRequests ({
 }) {
   /* The body is slight different for the initial and paginated requests */
   const formBody = {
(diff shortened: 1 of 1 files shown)


facundoolano:
Thanks. I'm seeing some `Unnecessary escape character: \"` errors, can you fix those?

tugbargn:
> Thanks. I'm seeing some `Unnecessary escape character: \"` errors, can you fix those?

I did't but i didn't work.

Here is the updated version (decoded version):
[REQUEST_TYPE.initial]: `f.req=[[["UsvDTd","[null,[2,${sort},[${numberOfReviewsPerRequest}],null,[null,null,null,null,null,null,null,null,2]],[\"${appId}\",7]]",null,"generic"]]]`,

PR #716, opened 2025-05-23
Add both commonjs and esm support

gjuoun:
Hi team

I believe there are still some projects are using commonjs,

this PR provides support for that

To verify, use `npm run build`

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+rm -rf dist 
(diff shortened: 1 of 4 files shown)


j-d-carmichael:
Did not see this... i went and built another solution to the same problem different approach though https://github.com/j-d-carmichael/google-play-scraper_cjs/blob/main/package.json#L24

PR #722, opened 2025-10-19
feat: esm & cjs output script for older projects that cannot afford to migrate to esm code.

j-d-carmichael:
ESM is nice, but we run a lot of older projects where the client just does not have budget to warrant a full migration to esm code.

This small addition keeps this project running smoothly on esm and commonJS projects while letting the source code enjoy the benefits of esm... best of both worlds.


j-d-carmichael:
The failing tests are not related to this PR

PR #752, opened 2026-05-21
Fix search function

ywachendorfer:
This PR should fix the broken search functionality by replacing the used endpoint /work/search with /store/search. Additionally, some tests had outdated expectations which lead to the issue that non related tests are failing on main branch. This tests where updated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^6.1.1",
     "eslint-plugin-standard": "^4.1.0",
-    "mocha": "^10.2.0",
+    "mocha": "^11.7.5",
     "promise-log": "^0.1.0",
     "sinon": "^15.2.0",
     "validator": "^13.9.0"
--- a/package.json
+++ b/package.json
@@ -24,25 +24,25 @@
   },
   "homepage": "https://github.com/facundoolano/google-play-scraper",
   "dependencies": {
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -40,7 +40,7 @@
     "eslint-plugin-node": "^11.1.0",
     "eslint-plugin-promise": "^6.1.1",
     "eslint-plugin-standard": "^4.1.0",
-    "mocha": "^10.2.0",
```
facundoolano, line 43: dont change dependencies as part of a separate change

ywachendorfer: Good point, i tried to fix the npm audit report which is failing because of mocha. However, even the newest version of mocha is not compliant with the github checks. I have reverted all changes regarding dependency updates

luca-regne:
hey, any update here? I've got the same error and the solution fixed the scrapping

deepakmkoshy:
Thank you, this solution fixed the search issue. Please merge it soon.

Kntntn:
I don't think it's an exact replacement, since the '/store/search' endpoint returns far fewer results than the '/work/search' for the keywords I tested. But I suppose it's better than nothing.

facundoolano:
> I don't think it's an exact replacement, since the '/store/search' endpoint returns far fewer results than the '/work/search' for the keywords I tested. But I suppose it's better than nothing.

I'm merging this one, if you have a better alternative please send a separate PR

PR #758, opened 2026-08-26
fix: correct updated timestamp and add published field (closes #212)

### LINKED ISSUE #212: updated does not match the date displayed in browser
The value of `updated` in the app details doesn't match the value displayed in browser. For example for the app used in the documentation (https://play.google.com/store/apps/details?id=com.dxco.pandavszombies&hl=en&gl=us) `May 27, 2015` is displayed in browser under `Updated`, though `updated` in the app details is set to `Feb 27, 2015`. What is the meaning of the latter date?

At the mapping path `'ds:3', 0, 12, 8, 0` seems to be the timestamp of the publish date. For the example app it is `1432677999` (2015-05-27 00:06:39).

aniruddhaadak80:
## What Problem This Solves

The `updated` field in app details returns the **release date** instead of the **actual update date** shown in the Play Store UI. For example, the app `com.dxco.pandavszombies` shows `May 27, 2015` in the browser under "Updated", but `updated` returns `Feb 27, 2015` (the release date).

## Why This Change Was Made

Per the maintainer's guidance in #212:
> "So we need to use that timestamp for the updated date, and the other for a new field, `published` date."

The timestamp at `'ds:3', 0, 12, 8, 0` corresponds to the actual update date displayed in the browser (e.g., `1432677999` = May 27, 2015).

## What Changed

- **`updated`**: Now uses path `['ds:3', 0, 12, 8, 0]` (actual update timestamp)
- **`published`** (new): Uses the previous `updated` path `['ds:5', 1, 2, 145, 0, 1, 0]` (release date)
- Both fields return milliseconds since epoch (same behavior as before)

## How Tested

- ESLint passes with no errors
- Existing test assertions (`assert.isNumber(app.updated)`) remain valid
- The fix aligns with the owner's explicit specification in #212

Fixes #212


aniruddhaadak80:
Diagnosis of the red `checks (16.x/18.x/20.x)` jobs (run 32983669925): not caused by this PR's diff.

All 84 mocha tests pass on every Node version. The failure is the subsequent `npm audit` step, which reports 6 pre-existing vulnerabilities in dev-dependency chains (`ajv`, `brace-expansion` via `mocha`, `diff`, `js-yaml`, `serialize-javascript`) and exits 1. This PR touches only `lib/app.js` (published/updated mapping) — no dependency changes — so the audit findings are base-branch drift, independent of this change.

Fixing it here would mean unrelated `npm audit fix` churn (including a breaking `mocha@8.1.3` downgrade note); that is a maintainer decision. Leaving the diff as-is.

facundoolano:
I'll merge with build passing and human generated description

PR #759, opened 2026-08-27
fix: don't crash when a developer page has no app list section

### LINKED ISSUE #730: Cannot read properties of undefined (reading 'fantasy-land/map')
Hello,

Some developer pages scrapping are failling with the following error:

```
TypeError: Cannot read properties of undefined (reading 'fantasy-land/map')
    at file:///home/node/app/node_modules/ramda/es/internal/_dispatchable.js:31:23
    at Module.f2 (file:///home/node/app/node_modules/ramda/es/internal/_curry2.js:28:14)
    at parseDeveloperApps (file:///home/node/app/node_modules/google-play-scraper/lib/developer.js:119:27)
    at file:///home/node/app/node_modules/google-play-scraper/lib/developer.js:52:29
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
```

It happens when calling the `developer` method on some developer accounts for instance: `7502834977667077022` [Store link](https://play.google.com/store/apps/dev?id=7502834977667077022&hl=en)

Here a code to reproduce this error:

```typescript
import scrapper from 'google-play-scraper';

const devIds: string[] = ['7502834977667077022'];

void (async () => {
  // eslint-disable-next-line no-restricted-syntax
  for (const devId of devIds) {
    try {
      await scrapper.developer({
        devId,
      });
    } catch (error) {
      console.log(`Fail to fetch ${devId}`, { error });
    }
  }
})();
```

Agi-Asi:
### Problem

`developer()` crashes on some developer pages:

```
TypeError: Cannot read properties of undefined (reading 'fantasy-land/map')
    at parseDeveloperApps (lib/developer.js:119)
```

Reproducible live with the developer id from #730:

```js
await gplay.developer({ devId: '7502834977667077022' });
```

The page (https://play.google.com/store/apps/dev?id=7502834977667077022) is a valid developer page ("Android Apps by Lategame Studio"), but its `ds:3` blob carries no app list section at the mapped path — the store renders that content lazily. `R.path()` returns `undefined`, and feeding that into `R.map()` produces the ramda crash above.

### Fix

Guard the extraction: when the app list section is missing, resolve with an empty list instead of crashing. Healthy pages are unaffected (verified live with both the string id `Jam City, Inc.` and the numeric id `5700313618786177705`).

### Testing

- New regression test against the exact developer id reported in #730 — crashes on `main`, passes with this fix.
- `npm test`: 85 passing, 0 failing
- `npm run lint`: clean

Fixes #730


Agi-Asi:

facundoolano:
- overcommented
- includes unrelated package changes
- dont add random developer dependency to tests

PR #760, opened 2026-08-27
fix: throw the intended error when an app has no similar-apps cluster

### LINKED ISSUE #701: Failed to fetch similar apps "Cannot read properties of null (reading 'length')"
* Operating System: Ubuntu24
* Node version: latest (also tested on node version 18)
* google-play-scraper version: latest release

### Description: 
i've use this library to fetch similar apps, but in some cases its cannot scrape similar apps

### Example code:
```js
  try {
    const similarApps = await gplay.similar({ appId, fullDetail: true });
    res.json(similarApps);
  } catch (error) {
    res.status(500).json({ error: 'Failed to fetch similar apps', details: error.message });
  }
```

### Error message:
```
{"error":"Failed to fetch similar apps","details":"Cannot read properties of null (reading 'length')"}
```

Agi-Asi:
### Problem

`similar()` fails with a `TypeError` instead of the intended domain error when an app page has no similar-apps cluster (region-restricted apps, apps without recommendations):

```
Cannot read properties of null (reading 'length')
```

as reported in #701. Root cause: `extractDataWithServiceRequestId` returns `null`/`undefined` when the `ag2B9c` service request section is absent from the parsed page, and `parseSimilarApps` reads `.length` on that before the intended `Error('Similar apps not found')` can fire. Callers can't distinguish "this app has no similar apps" from a parser bug.

### Fix

- `parseSimilarApps`: accept only a non-empty array of clusters; anything else throws the domain error `Similar apps not found`.
- `processFirstPage` (same file): the same guard for the cluster page's app list section — its mapping feeds `R.map`, which crashes the same way on `undefined` input.

`parseSimilarApps` is additionally exported (named export; the default export is unchanged) so the guard can be pinned with a synthetic payload — which live apps happen to lack a cluster changes over time, so a live-repro test would be flaky.

### Testing

- New unit test: a parsed payload without the `ag2B9c` section throws `/Similar apps not found/` instead of a `TypeError`.
- Live behavior unchanged for normal apps (`com.mojang.minecraftpe`, `com.spotify.music`).
- `npm test`: 85 passing, 0 failing; `npm run lint`: clean

Fixes #701


Agi-Asi:

PR #761, opened 2026-08-27
fix: categories() returned only ['APPLICATION'] — serve the canonical id list

### LINKED ISSUE #671: categories() does not list all the categories.
* Operating System: Windows
* Node version: 18.17.0
* google-play-scraper version: ^10.0.0

### Description: 
when using categories(), It returns a list with only one item.

### Example code:
```js
import gplay from 'google-play-scraper';
const categories = await gplay.categories();
console.log(categories);
```

### Output:

```
['APPLICATION']
```

Agi-Asi:
### Problem

`categories()` returns just `['APPLICATION']` (#671). Google Play no longer renders the category navigation menu in the static HTML of `/store/apps` — it is lazy-loaded client side — so the cheerio scrape of `ul li a` hrefs finds nothing and only the hardcoded `APPLICATION` entry survives. Verified live: the store page markup contains **zero** matching anchors today (`$('ul li a')` → 0 elements).

### Fix

Serve the category ids from `constants.category` instead of scraping. That list is already the source of truth this library maintains (it's what `list()` accepts), and every id maps to a live `/store/apps/category/<ID>` page — spot-checked `APPLICATION`, `GAME`, `GAME_ACTION`, `ART_AND_DESIGN`, `WATCH_FACE`, `ANDROID_WEAR`, `FAMILY` against the live store, all 200s.

The function keeps its promise-returning signature and keeps accepting (and ignoring) its options argument, so existing callers are unaffected — they just get the full 54-entry list back instead of one entry.

Notably, the existing test `should have all categories from constant list of categories` asserts exactly `categories() ⊆ constants.category`, which this implementation satisfies by construction.

### Testing

- Existing categories tests pass; the first one gains a regression guard against the single-entry degradation (`length > 1`, includes `APPLICATION` and `GAME`).
- `npm test`: 84 passing, 0 failing; `npm run lint`: clean

Fixes #671


Agi-Asi:

